PR #25, opened 2022-11-02
Container images

### LINKED ISSUE #23: Containerfiles and GHCR deployments

DanNixon:
Closes #23.


.github/workflows/ci.yml
```diff
@@ -42,3 +44,40 @@ jobs:
 
       - name: Build and Test
         run: cargo test --all
+
+  container-images:
+    name: Build and push container image
+    if: ${{ github.ref_type == 'tag' }}
```
Pasarus, line 50: Curious as to why you are using this methodology for limiting when a build and push of container images occurs?

Is it likely you will want to force a build on specifically PRs? If not and you will always be building the main branch you could consider using a
```
on:
  "workflow_dispatch"
```
Which allows you to just hit a button in the actions tab for this workflow.

This would also probably required you to move it to a new .yml workflow file.

Pasarus (changes requested):
Just a quick question, other than that, it looks good

.github/workflows/ci.yml
DanNixon: > Curious as to why you are using this methodology for limiting when a build and push of container images occurs?

- I want the main CI to run before building the container images
- I do not want to redefine the main CI in another workflow
- The container images should only ever build for tags

If there is an alternative way to do this that is cleaner I'm open to suggestions, I've probably not exhausted all you can do with GH Actions.

> Is it likely you will want to force a build on specifically PRs?

Probably not.

> If not and you will always be building the main branch you could consider using a

Not sure how this helps, that seems to only be mentioned as a way of running jobs manually, which is not desired.

DanNixon: Although maybe being able to get a container build for a specific component for a specific Git SHA (or head of a specific branch) would be a nice to have.

I can imagine that would be good for testing in cluster.

PR #33, opened 2022-11-17
Prevent data corruption when a digitizer does not send some data

### LINKED ISSUE #30: stream-to-file: Ensure temporary loss in digitizer data does not break files

DanNixon:
Code review:
 - See that the `multiple_digitizers_missing_data` test accurately tests the failure case.
 - See that tests were not modified in the test cleanup commit.

Fixes #30.


stream-to-file/src/file/trace/tests/multiple_digitizers_missing_data.rs
```diff
@@ -0,0 +1,175 @@
...
+    push_frame(
+        &mut file,
+        num_time_points,
+        0,
+        GpsTime::new(22, 205, 10, 55, 30, 0, 0, 0),
+        0,
+        1,
+    );
+
+    // push_frame(
```
Pasarus, line 54: Could we delete (or implement) these commented out tests?

stream-to-file/src/file/trace/tests/multiple_digitizers_missing_data.rs
```diff
@@ -0,0 +1,175 @@
...
+    push_frame(
+        &mut file,
+        num_time_points,
+        2,
+        GpsTime::new(22, 205, 10, 55, 30, 40, 0, 0),
+        0,
+        1,
+    );
+
+    // push_frame(
```
Pasarus, line 72: Could we delete (or implement) these commented out tests?

Pasarus (changes requested):
Just a question about some commented out code

stream-to-file/src/file/trace/tests/multiple_digitizers_missing_data.rs
DanNixon: I'm inclined to leave them as they are, the test data in this test is a copy of the normal case multiple detectors test but with some missing data, this is the missing data.

Pasarus: Could you comment as to why this is a comment and not actually used in the test?

Pasarus (approved):

PR #40, opened 2023-05-16
Trace to HDF5 file archiver

DanNixon:
See the added README.md for a general description of this tool.


Modularius (commented):
Would it be a good idea to include the sample rate in the metadata saved to the h5 file?

DanNixon:
> Would it be a good idea to include the sample rate in the metadata saved to the h5 file?

Yes, I think I even considered that and have no idea why I did not include it.

Modularius (approved):

PR #48, opened 2023-08-15
Unauthenticated Kafka Connections

henrydenny246:
Allows the `username` and `password` command line arguments to be omitted, in which case an unauthenticated Kafka connection is able to be created instead.

### DIFF
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -23,7 +23,7 @@ pub fn channel_index(digitizer_index: usize, channel_index: usize) -> usize {
     (digitizer_index * CHANNELS_PER_DIGITIZER) + channel_index
 }
 
-pub fn generate_client_config(
+pub fn generate_kafka_client_config(
     broker_address: &String,
     username: &Option<String>,
     password: &Option<String>,
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -25,7 +25,7 @@ pub fn channel_index(digitizer_index: usize, channel_index: usize) -> usize {
 
 pub fn generate_client_config(
     broker_address: &String,
-    opt_username: &Option<String>,
+    username: &Option<String>,
     opt_password: &Option<String>,
 ) -> ClientConfig {
     let mut client_config = ClientConfig::new()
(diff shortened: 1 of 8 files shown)


DanNixon:
Looks like you need to rebase this, run `git fetch` followed by `git rebase origin/main`.

henrydenny246:
I've run the commands but can't see a difference. Has it worked?

common/src/lib.rs
```diff
@@ -20,3 +22,23 @@ pub const CHANNELS_PER_DIGITIZER: usize = 8;
 pub fn channel_index(digitizer_index: usize, channel_index: usize) -> usize {
     (digitizer_index * CHANNELS_PER_DIGITIZER) + channel_index
 }
+
+pub fn generate_client_config(
+    broker_address: &String,
+    opt_username: &Option<String>,
```
DanNixon, line 28:
```suggestion
    username: &Option<String>,
```

common/src/lib.rs
```diff
@@ -20,3 +22,23 @@ pub const CHANNELS_PER_DIGITIZER: usize = 8;
 pub fn channel_index(digitizer_index: usize, channel_index: usize) -> usize {
     (digitizer_index * CHANNELS_PER_DIGITIZER) + channel_index
 }
+
+pub fn generate_client_config(
```
DanNixon, line 26:
```suggestion
pub fn generate_kafka_client_config(
```

common/src/lib.rs
```diff
@@ -8,6 +8,8 @@ pub type Intensity = u16;
 pub type FrameNumber = u32;
 pub type SampleRate = u64;
 
+use rdkafka::config::ClientConfig;
```
DanNixon, line 11: Could this be moved up so that it is between the end of modules and start of types, this is the general format used elsewhere in the codebase.

DanNixon (changes requested):
Could the new function also be used in the other binaries that handle Kafka connections?

DanNixon (approved):
:+1:

PR #54, opened 2023-08-31
Kafka DAQ Report

Henry-Denny:


kafka-daq-report/src/main.rs
```diff
@@ -0,0 +1,294 @@
...
+    password: Option<String>,
+
+    #[clap(long = "group")]
+    consumer_group: String,
+
+    #[clap(long)]
+    trace_topic: String,
+
+    #[clap(long, default_value = ".")]
+    output: PathBuf,
```
DanNixon, line 92: Is this option required?

kafka-daq-report/src/main.rs
```diff
@@ -0,0 +1,294 @@
...
+
+    #[clap(long, default_value = "127.0.0.1:9090")]
+    observability_address: SocketAddr,
+}
+
+enum Event<I> {
+    Input(I),
+    Tick,
```
DanNixon, line 100: Is the `Tick` event used?

kafka-daq-report/src/ui.rs
```diff
@@ -0,0 +1,104 @@
...
+                Constraint::Length(3),
+                Constraint::Min(1),
+                Constraint::Length(3),
+            ]
+            .as_ref(),
+        )
+        .split(frame.size());
+
+    // Draw all widgets.
+    draw_title(frame, chunks[0]);
```
DanNixon, line 26: For the sake of saving screen space, the title banner is probably not required.

DanNixon (changes requested):
Looks good :+1:.

Just a few small comments.

kafka-daq-report/src/main.rs
Henry-Denny: I copied the event polling algorithm from an online example, and I'm not entirely sure how it works. The TUI does not get displayed if I remove the `Tick` event.

kafka-daq-report/src/main.rs
Henry-Denny: No, I have removed it.

kafka-daq-report/src/main.rs
```diff
@@ -0,0 +1,294 @@
...
+
+use anyhow::Result;
+use chrono::{DateTime, Utc};
+use clap::Parser;
+use crossterm::event::{self, DisableMouseCapture, EnableMouseCapture, Event as CEvent, KeyCode};
+use crossterm::execute;
+use crossterm::terminal::{
+    disable_raw_mode, enable_raw_mode, EnterAlternateScreen, LeaveAlternateScreen,
+};
+use kagiyama::{AlwaysReady, Watcher};
```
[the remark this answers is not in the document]
Henry-Denny: Would you mind clarifying what you mean by the 'observability code'?

kafka-daq-report/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[workspaces]
+resolver = "2"
```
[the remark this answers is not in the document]
Henry-Denny: I added this line because I kept getting warnings in the terminal about there being a mismatch of versions. It did not end up having any effect but I forgot to remove it.

kafka-daq-report/src/main.rs
DanNixon: Ah, I see how it is being used in the example now.

And it has just occurred to me that the periodic update is actually required here, my suggestion was going to be a refactor that would end up meaning you would have to move the cursor to see any value updating.

kafka-daq-report/src/main.rs
```diff
@@ -0,0 +1,285 @@
...
+    password: Option<String>,
+
+    #[clap(long = "group")]
+    consumer_group: String,
+
+    #[clap(long)]
+    trace_topic: String,
+
+    #[clap(long, default_value = "127.0.0.1:9090")]
+    observability_address: SocketAddr,
```
DanNixon, line 89: This should also be removed (I actually thought you'd get a warning about it being unused after removing the other observability code).

DanNixon (approved):

PR #91, opened 2023-11-30
Inherit Dependencies

Henry-Denny:
The `common` and `streaming-types` dependencies are now inherited from the main `Cargo.toml` file.


Cargo.toml
```diff
@@ -17,6 +17,9 @@ license = "GPL-3.0-only"
 edition = "2021"
 
 [workspace.dependencies]
+common = { path = "./common" }
```
DanNixon, line 20:
```suggestion
supermusr-common = { path = "./common" }
```

Prefixing these would be a good idea to make it perfectly clear what they are.

Cargo.toml
```diff
@@ -17,6 +17,9 @@ license = "GPL-3.0-only"
...
+streaming-types = { path = "./streaming-types" }
```
DanNixon, line 21:
```suggestion
supermusr-streaming-types = { path = "./streaming-types" }
```

Cargo.toml
```diff
@@ -17,6 +17,9 @@ license = "GPL-3.0-only"
...
+
```
DanNixon, line 22:
```suggestion
```

Remove this newline and keep the dependency list sorted alphabetically.

Cargo.toml
```diff
@@ -27,6 +30,11 @@ lazy_static = "1.4.0"
 log = "0.4"
 ndarray = "0.15.6"
 ndarray-stats = "0.5.1"
-rdkafka = { version = "0.31.0", features = ["cmake-build", "ssl", "gssapi", "sasl"] }
+rdkafka = { version = "0.31.0", features = [
```
DanNixon, line 33: Was there a reason for this? It would be preferable to keep the one line per dependency format.

DanNixon (changes requested):
There are also some crates that still have dependency versions in their `Cargo.toml`.

Cargo.toml
Henry-Denny: This was automatically formatted when I saved the file for some reason. I'm not sure whether it would be the LSP doing this or `treefmt`.

Henry-Denny:
> There are also some crates that still have dependency versions in their `Cargo.toml`.

Are you referring to the crates that are only included in one place like `itertools`?

DanNixon:
> > There are also some crates that still have dependency versions in their `Cargo.toml`.
> 
> Are you referring to the crates that are only included in one place like `itertools`?

Yes.

Cargo.toml
DanNixon: You'll also need to rename these in the crates that inherit the dependency.

Henry-Denny: Sorry, I've just noticed that as well.

DanNixon (changes requested):

docs/style.md
```diff
@@ -27,9 +27,9 @@ instead of
 use crate::Something;
 
 use tokio::task::JoinHandle;
-use supermusr_common::Time
-
 use std::time::Duration
+
+use supermusr_common::Time
```
DanNixon, line 32: I only meant you to sort the example above, this snippet is supposed to be an example of what not to do.

DanNixon (changes requested):

PR #92, opened 2023-12-01
Kafka DAQ Report Metrics

### LINKED ISSUE #90: kafka-daq-report metrics
- Change the timestamps to include date as well as time
- Add message rate in Hz

Henry-Denny:
Closes #90.

### DIFF
--- a/kafka-daq-report/src/main.rs
+++ b/kafka-daq-report/src/main.rs
@@ -34,6 +34,7 @@ pub struct DigitiserData {
     pub first_msg_timestamp: Option<DateTime<Utc>>,
     pub last_msg_timestamp: Option<DateTime<Utc>>,
     pub last_msg_frame: u32,
+    pub message_rate: u64,
     pub num_channels_present: usize,
     pub has_num_channels_changed: bool,
     pub num_samples_in_first_channel: usize,
@@ -46,6 +47,7 @@ impl DigitiserData {
     pub fn new(
         timestamp: Option<DateTime<Utc>>,
         frame: u32,
+        message_rate: u64,
         num_channels_present: usize,
         num_samples_in_first_channel: usize,
         is_num_samples_identical: bool,
@@ -55,6 +57,7 @@ impl DigitiserData {
             first_msg_timestamp: timestamp,
             last_msg_timestamp: timestamp,
             last_msg_frame: frame,
+            message_rate,
             num_channels_present,
             has_num_channels_changed: false,
             num_samples_in_first_channel,
@@ -202,6 +205,7 @@ async fn poll_kafka_msg(consumer: StreamConsumer, shared_data: SharedData) {
                                 let mut logged_data = shared_data.lock().unwrap();
 
                                 let frame_number = data.metadata().frame_number();
+                                let message_rate = data.sample_rate();
 
                                 let num_channels_present = match data.channels() {
                                     Some(c) => c.len(),
@@ -234,6 +238,7 @@ async fn poll_kafka_msg(consumer: StreamConsumer, shared_data: SharedData) {
                                         d.num_msg_received += 1;
 
                                         d.last_msg_timestamp = timestamp;
+                                        d.message_rate = message_rate;
                                         let num_channels = match data.channels() {
                                             Some(c) => c.len(),
                                             None => 0,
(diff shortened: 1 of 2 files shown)


kafka-daq-report/src/main.rs
```diff
@@ -238,7 +236,16 @@ async fn poll_kafka_msg(consumer: StreamConsumer, shared_data: SharedData) {
                                         d.num_msg_received += 1;
 
                                         d.last_msg_timestamp = timestamp;
-                                        d.message_rate = message_rate;
+
+                                        d.message_rate = if let (Some(first), Some(last)) =
```
DanNixon, line 240: Could a message count over a fixed time window be used instead.

This approach would have issues responding to and accurately representing unstable message receive rates, plus reliance on the timestamps should be avoided.

DanNixon (changes requested):

kafka-daq-report/src/main.rs
Henry-Denny: Yes, that makes sense. Should the fixed time window be hard-coded in or passed in as a command line argument?

DanNixon: A command line argument with a default of 5 seconds would be good.

DanNixon:
@Henry-Denny Is this still being worked on or is it ready for another review?

Henry-Denny:
@DanNixon It is ready for another review.

DanNixon:
Is the issue you mentioned regarding not updating after data is not received fixed?

kafka-daq-report/src/main.rs
```diff
@@ -228,36 +242,55 @@ async fn poll_kafka_msg(consumer: StreamConsumer, shared_data: SharedData) {
...
+                                            timestamp,
+                                            frame_number,
+                                            num_channels_present,
+                                            num_samples_in_first_channel,
+                                            is_num_samples_identical,
+                                        ));
+                                };
+
+                                let shared_data = Arc::clone(&shared_data);
+                                thread::spawn(move || {
```
DanNixon, line 287: Is there a reason for using `std::thread` instead of `tokio::task`?

kafka-daq-report/src/main.rs
```diff
@@ -228,36 +242,55 @@ async fn poll_kafka_msg(consumer: StreamConsumer, shared_data: SharedData) {
...
+                                    thread::sleep(Duration::from_secs(recent_message_lifetime));
```
DanNixon, line 288: This appears to only compute the message rate after a given delay of a message being received.
This is not the intended behaviour, the message rate should be calculated at a constant rate (defined by `recent_message_lifetime` here) regardless of what messages are being received.
This is why you are seeing issues when no more messages are received.

DanNixon (changes requested):

kafka-daq-report/src/main.rs
Henry-Denny: I set the code up to track the number of messages detected in the last `recent_message_lifetime` seconds, increasing the count whenever a message is detected and decreasing the count after the set period of time. This way the `message_rate` can be calculated dynamically every frame. I'm not sure why this approach doesn't work. Surely the message rate must depend on the messages being received?

kafka-daq-report/src/main.rs
Henry-Denny: I was better able to understand the documentation for `std::thread` than I was for `tokio::task` so I opted on the using the former. I will switch to using `tokio::task` instead.

kafka-daq-report/src/main.rs
DanNixon: The two issues with this are:
- You do not have a constant sampling window, so the value displayed may fluctuate needlessly.
- It does not work if the message rate is below the sampling rate (which should not typically happen, but this is a diagnostics tool, so should be able to cope with things not working as intended).
- (a more technical point is that spawning a task or (especially) thread for each received message is needlessly expensive)

As number of messages received is already being recorded, all that you need to do is evey n seconds calculate the difference in this number since the last iteration, divide it by n and you have the message rate in Hz.

Henry-Denny: Ok, I will use that approach instead. That should mean that the `message_rate` is only updated every `recent_message_lifetime` seconds but I assume that is not an issue?

DanNixon: > Ok, I will use that approach instead. That should mean that the `message_rate` is only updated every `recent_message_lifetime` seconds but I assume that is not an issue?

That is the intention, although the parameter should be renamed to better reflect this.

DanNixon (approved):

PR #104, opened 2024-01-30
Run simulator

Modularius:
Generates RunStart and RunStop messages for the data pipeline:
To generate a RunStart message run

"run-simulator --broker localhost:19092 --topic Control --run-name Test --time "2024-01-30 15:17:03.618842621Z" run-start --instrument-name SuperMuSR"

To generate a RunStop message run
"run-simulator --broker localhost:19092 --topic Control --run-name Test --time "2024-01-30 15:17:03.618842621Z" run-stop"

Note the run-name should be the same for corresponding start/stop messages, time can be omitted if desired (the present time is used), and instrument-name is used by the nexus writer downstream.


DanNixon:
This looks good.

I do however think this would be better suited as an additional mode on the `simulator` (i.e. when invoked it can emit e.g. the test event pattern, or simulated data with a given function, or a run start, etc...).

Would this be much work to do now given you were discussing doing this for the original simulator and your function based simulator?

Modularius:
It would be easy enough to do that, however it would make the simulator tool more complex to use, and increase the chance of run message simulation being affected by future changes I plan to make in the simulator tool. It might be appropriate to fold the run simulator into the simulator eventually but for the moment it might be better to keep them separate.

DanNixon:
Ok, can you open an issue describing the work you have planned for consolidating the various simulators into a single tool.

Modularius:
Yep, sure

DanNixon (approved):

PR #108, opened 2024-02-09
Send 8 channels when simulating patterned trace data

DanNixon:


Modularius:
I'm working on an updated simulator that allows you to send complex groups of messages though settings in a JSON file. I should be able to handle this updated functionality.

Is this PR because you need an 8 channel simulator right now, as my update won't be ready for a couple of days?

DanNixon:
I was not aware the simulator rewrite you had in mind was the extensive, as far as I am concerned what we need is basically something to cover two use cases: testing throughput and testing event detection, this PR addresses the former and currently the trace reader does the latter.

While I am not against having a very versatile arbitrary waveform generator, there are other things that are a higher priority right now.

PR #111, opened 2024-02-14
Open metrics digitiser statistics

### LINKED ISSUE #95: Open metrics digitiser statistics
A version of `kafka-daq-report` that reports digitiser statistics via OpenMetrics using https://metrics.rs/.

Metrics to be reported and labels they should carry:
- "digitiser_message_received_count" (dig. ID) : Number of messages received
- "digitiser_last_message_timestamp" (dig. ID) : Timestamp of last message received (in ns unix epoch)
- "digitiser_last_message_frame_number" (dig. ID) : Frame number of last message received
- "digitiser_channel_count" (dig. ID): Number of channels in last message received
- "digitiser_sample_count" (dig. ID, channel number): Number of samples in last message received

Henry-Denny:
Closes #95


open-metrics-report/src/main.rs
```diff
@@ -0,0 +1,201 @@
...
+
+                                // Channel count
+                                let channel_count = match data.channels() {
+                                    Some(c) => c.len(),
+                                    None => 0,
+                                };
+                                gauge!("digitiser_channel_count", &labels)
+                                    .set(channel_count as f64);
+
+                                // Sample count
```
DanNixon, line 161: Sample count also has a label of channel number, so all channels should be reported.

open-metrics-report/src/main.rs
```diff
@@ -0,0 +1,201 @@
...
+                                    data.metadata().timestamp().copied().map(|t| t.into());
+                                gauge!("digitiser_last_message_timestamp", &labels)
+                                    .set(timestamp.unwrap().nanosecond() as f64);
+
+                                /* Logging */
+                                log::info!(
+                                    "Trace packet: dig. ID: {}, metadata: {:?}",
+                                    data.digitizer_id(),
+                                    data.metadata()
+                                );
```
DanNixon, line 180:
```suggestion
```

open-metrics-report/src/main.rs
```diff
@@ -0,0 +1,201 @@
...
+                                    None => 0,
+                                };
+                                gauge!("digitiser_sample_count", &labels)
+                                    .set(num_samples_in_first_channel as f64);
+
+                                // Last message timestamp
+                                let timestamp: Option<GpsTime> =
+                                    data.metadata().timestamp().copied().map(|t| t.into());
+                                gauge!("digitiser_last_message_timestamp", &labels)
+                                    .set(timestamp.unwrap().nanosecond() as f64);
```
DanNixon, line 173: What does `nanosecond()` actually return in this case?

DanNixon (changes requested):
Could this tool also be renamed `trace-telemetry-adapter`, this better describes the purpose of the tool.

open-metrics-report/src/main.rs
Henry-Denny: Just realised that this will be the nanosecond part of the timestamp (omitting the date, hours and minutes). Should this be the number of seconds since the Unix Epoch instead?

DanNixon: Unix timestamp would be the best (and pretty much only in OpenMetrics) option for this, it would need nanosecond precision however. Representing this as a `f64` with the unit as seconds should be good enough.

DanNixon:
Please revert your changes to `flake.lock`.

trace-telemetry-adapter/src/main.rs
```diff
@@ -0,0 +1,195 @@
...
+use metrics_exporter_prometheus::PrometheusBuilder;
+use rdkafka::{
+    consumer::{stream_consumer::StreamConsumer, CommitMode, Consumer},
+    message::Message,
+};
+use supermusr_streaming_types::dat1_digitizer_analog_trace_v1_generated::{
+    digitizer_analog_trace_message_buffer_has_identifier, root_as_digitizer_analog_trace_message,
+};
+
+/*
```
DanNixon, line 16: This comment is not required assuming the changes to `README.md`.

DanNixon:
> Could this tool also be renamed `trace-telemetry-adapter`, this better describes the purpose of the tool.

Actually, on second thought could the name be `trace-telemetry-exporter`, this is more in-keeping with the observability ecosystem it fits into.

trace-telemetry-exporter/src/main.rs
```diff
@@ -0,0 +1,185 @@
+use std::net::SocketAddr;
+
```
DanNixon, line 2:
```suggestion
```

trace-telemetry-exporter/README.md
```diff
@@ -0,0 +1,19 @@
+# trace-telemetry-exporter
+
+A tool which uses the OpenMetrics format to record the following:
```
DanNixon, line 3: Could this be formatted in a tabular format, similar to https://github.com/charlie-haley/omada_exporter?tab=readme-ov-file#-metrics

DanNixon (changes requested):

trace-telemetry-exporter/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+| Name                                | Description                                                   | Labels                         |
+|-------------------------------------|---------------------------------------------------------------|--------------------------------|
+| digitiser_message_received_count    | The number of messages received by a digitiser.               | `digitiser_id`                 |
+| digitiser_last_message_timestamp    | The timestamp of the last message received by a digitiser.    | `digitiser_id`                 |
```
DanNixon, line 8: Include details of what this actually is (i.e. unit and epoch).

trace-telemetry-exporter/README.md
```diff
@@ -0,0 +1,11 @@
+# trace-telemetry-exporter
+
+A tool which uses the OpenMetrics format to record the following:
+
+| Name                                | Description                                                   | Labels                         |
+|-------------------------------------|---------------------------------------------------------------|--------------------------------|
+| digitiser_message_received_count    | The number of messages received by a digitiser.               | `digitiser_id`                 |
```
DanNixon, line 7:
```suggestion
| digitiser_message_received_count    | The number of messages received from a digitiser.               | `digitiser_id`                 |
```

trace-telemetry-exporter/README.md
```diff
@@ -0,0 +1,11 @@
...
+| digitiser_last_message_timestamp    | The timestamp of the last message received by a digitiser.    | `digitiser_id`                 |
+| digitiser_last_message_frame_number | The frame number of the last message received by a digitiser. | `digitiser_id`                 |
+| digitiser_channel_count             | The number of channels in a digitiser.                        | `digitiser_id`                 |
```
DanNixon, line 10: Associate this to the last message received, not just in general as suggested in this description.

trace-telemetry-exporter/README.md
```diff
@@ -0,0 +1,11 @@
...
+| digitiser_sample_count              | The number of samples in a channel of a digitiser             | `digitiser_id` `channel_index` |
```
DanNixon, line 11: Same as above comment.

trace-telemetry-exporter/src/main.rs
```diff
@@ -0,0 +1,184 @@
...
+    password: Option<String>,
+
+    #[clap(long = "group")]
+    consumer_group: String,
+
+    #[clap(long)]
+    trace_topic: String,
+
+    #[clap(long, default_value = "127.0.0.1:9090")]
+    metrics_address: SocketAddr,
```
DanNixon, line 40: Allow this to also be set by environment variable (similar to `OBSERVABILITY_ADDRESS` elsewhere).

trace-telemetry-exporter/default.nix
```diff
@@ -0,0 +1,42 @@
...
+    tag = "latest";
+    created = "now";
+
+    copyToRoot = pkgs.buildEnv {
+      name = "image-root";
+      paths = with pkgs; [bashInteractive coreutils];
+      pathsToLink = ["/bin"];
+    };
+
+    config = {
```
DanNixon, line 35: Set the metrics address variable here, to `0.0.0.0:9090`.

trace-telemetry-exporter/src/main.rs
```diff
@@ -0,0 +1,184 @@
...
+                                    }
+                                }
+
+                                // Last message timestamp
+                                let timestamp: DateTime<Utc> =
+                                    data.metadata().timestamp().copied().unwrap().into();
+                                gauge!("digitiser_last_message_timestamp", &labels)
+                                    .set(timestamp.timestamp_nanos_opt().unwrap() as f64);
+
+                                log::info!(
```
DanNixon, line 165:
```suggestion
                                log::debug!(
```

trace-telemetry-exporter/src/main.rs
```diff
@@ -0,0 +1,184 @@
...
+    Ok(())
+}
+
+/// Poll kafka messages and update digitiser data.
+async fn poll_kafka_msg(consumer: StreamConsumer) {
+    loop {
+        match consumer.recv().await {
+            Err(e) => log::warn!("Kafka error: {}", e),
+            Ok(msg) => {
+                log::debug!(
```
DanNixon, line 110:
```suggestion
                log::trace!(
```

trace-telemetry-exporter/src/main.rs
```diff
@@ -0,0 +1,184 @@
...
+        "Number of samples in last message received"
+    );
+
+    poll_kafka_msg(consumer).await;
+
+    Ok(())
+}
+
+/// Poll kafka messages and update digitiser data.
+async fn poll_kafka_msg(consumer: StreamConsumer) {
```
DanNixon, line 105: There is a lot of nesting in this function, could you move the section that actually processes the message (i.e. line 123 onwards) to a separate function.

DanNixon (changes requested):

trace-telemetry-exporter/src/main.rs
DanNixon: Yes, that is expected.

Henry-Denny: I'm a bit confused. I thought that `OBSERVABILITY_ADDRESS` could be set by environment variable?

DanNixon: It can, but only when configured to do so. Just checking and this does seem to have been omitted elsewhere, could you add the correct flag for these (see https://github.com/DanNixon/satori/blob/main/event-processor/src/main.rs#L30).

trace-telemetry-exporter/src/main.rs
```diff
@@ -0,0 +1,185 @@
...
+    #[clap(long)]
+    password: Option<String>,
+
+    #[clap(long = "group")]
+    consumer_group: String,
+
+    #[clap(long)]
+    trace_topic: String,
+
+    #[clap(long, env, default_value = "127.0.0.1:9090")]
```
DanNixon, line 41:
```suggestion
    #[clap(long, env, default_value = "127.0.0.1:9091")]
```

trace-telemetry-exporter/default.nix
```diff
@@ -0,0 +1,43 @@
...
+      Entrypoint = ["${pkgs.tini}/bin/tini" "--" "${trace-telemetry-exporter}/bin/trace-telemetry-exporter"];
+      Env = [
+        "SSL_CERT_FILE=${pkgs.cacert}/etc/ssl/certs/ca-bundle.crt"
+        "METRICS_ADDRESS=0.0.0.0:9090"
```
DanNixon, line 39:
```suggestion
        "METRICS_ADDRESS=0.0.0.0:9091"
```

DanNixon (changes requested):

trace-telemetry-exporter/src/main.rs
```diff
@@ -0,0 +1,185 @@
...
+        .with_http_listener(args.metrics_address)
+        .install()
+        .expect("prometheus metrics exporter should be setup");
+
+    metrics::describe_counter!(
+        METRIC_MSG_COUNT,
+        metrics::Unit::Count,
+        "Number of messages received"
+    );
+
```
DanNixon, line 78: Just realised I missed an obvious metric out of the initial list.

Message rate should also be included.

DanNixon (changes requested):

trace-to-events/src/main.rs
```diff
@@ -38,7 +38,7 @@ struct Cli {
     #[clap(long)]
     event_topic: String,
 
-    #[clap(long, default_value = "127.0.0.1:9090")]
+    #[clap(long, env, default_value = "127.0.0.1:9090")]
```
DanNixon, line 41: Out of scope for this PR.

trace-telemetry-exporter/src/main.rs
```diff
@@ -0,0 +1,228 @@
...
+    ("digitiser_id".to_string(), format!("{}", digitiser_id))
+}
+
+async fn update_message_rate(recent_msg_counts: MessageCounts, recent_message_lifetime: u64) {
+    loop {
+        // Wait a set period of time before calculating average.
+        sleep(Duration::from_secs(recent_message_lifetime)).await;
+        let mut recent_msg_counts = recent_msg_counts.lock().unwrap();
+        // Calculate and record message rate for each digitiser.
+        recent_msg_counts.iter().for_each(|recent_msg_count| {
```
DanNixon, line 140: You should be able to expand this in the argument list.
```suggestion
        recent_msg_counts.iter().for_each(|(digitiser_id, recent_msg_count)| {
```

trace-telemetry-exporter/src/main.rs
```diff
@@ -0,0 +1,228 @@
...
+    poll_kafka_msg(consumer).await;
+
+    Ok(())
+}
+
+fn get_digitiser_label(digitiser_id: u8) -> (String, String) {
+    ("digitiser_id".to_string(), format!("{}", digitiser_id))
+}
+
+async fn update_message_rate(recent_msg_counts: MessageCounts, recent_message_lifetime: u64) {
```
DanNixon, line 134: The name of the second argument is misleading, suggest changing it to match the command line flag name.

trace-telemetry-exporter/src/main.rs
```diff
@@ -0,0 +1,228 @@
...
+        "Number of channels in last message received"
+    );
+
+    metrics::describe_gauge!(
+        METRIC_SAMPLE_COUNT,
+        metrics::Unit::Count,
+        "Number of samples in last message received"
+    );
+
+    let recent_msg_counts: MessageCounts = Arc::new(Mutex::new(HashMap::new()));
```
DanNixon, line 118: Where is the code that updates these counts?

DanNixon (changes requested):

trace-telemetry-exporter/src/main.rs
```diff
@@ -0,0 +1,236 @@
...
+
+    task::spawn(update_message_rate(
+        Arc::clone(&recent_msg_counts),
```
DanNixon, line 121:
```suggestion
        recent_msg_counts.clone(),
```

trace-telemetry-exporter/src/main.rs
```diff
@@ -0,0 +1,236 @@
...
+}
+
+fn get_digitiser_label(digitiser_id: u8) -> (String, String) {
+    ("digitiser_id".to_string(), format!("{}", digitiser_id))
+}
+
+async fn update_message_rate(recent_msg_counts: MessageCounts, message_rate_interval: u64) {
+    loop {
+        // Wait a set period of time before calculating average.
+        sleep(Duration::from_secs(message_rate_interval)).await;
```
DanNixon, line 137:
```suggestion
        sleep(Duration::from_secs(message_rate_interval)).await;
        
```

trace-telemetry-exporter/src/main.rs
```diff
@@ -0,0 +1,236 @@
...
+    );
+
+    let recent_msg_counts: MessageCounts = Arc::new(Mutex::new(HashMap::new()));
+
+    task::spawn(update_message_rate(
+        Arc::clone(&recent_msg_counts),
+        args.message_rate_interval,
+    ));
+
+    poll_kafka_msg(consumer, Arc::clone(&recent_msg_counts)).await;
```
DanNixon, line 125:
```suggestion
    poll_kafka_msg(consumer, recent_msg_counts).await;
```

trace-telemetry-exporter/src/main.rs
```diff
@@ -0,0 +1,236 @@
...
+        sleep(Duration::from_secs(message_rate_interval)).await;
+        let mut recent_msg_counts = recent_msg_counts.lock().unwrap();
+        // Calculate and record message rate for each digitiser.
+        recent_msg_counts
+            .iter()
+            .for_each(|(digitiser_id, recent_msg_count)| {
+                let msg_rate = *recent_msg_count as f64 / message_rate_interval as f64;
+                let labels = [get_digitiser_label(*digitiser_id)];
+                gauge!("digitiser_message_received_rate", &labels).set(msg_rate);
+            });
```
DanNixon, line 146:
```suggestion
            });
            
```

trace-telemetry-exporter/src/main.rs
```diff
@@ -0,0 +1,236 @@
...
+        recent_msg_counts.clear();
+    }
+}
+
+fn process_message(data: &DigitizerAnalogTraceMessage<'_>, recent_msg_counts: MessageCounts) {
+    let id = data.digitizer_id();
+    let labels = [get_digitiser_label(id)];
+
+    // Increment recent message count for the digitiser
+    recent_msg_counts
```
DanNixon, line 157: Has this been tested? Pretty sure this will result in no entries being added to the recent counts as you are not handling the case where they key does not exist. (I admit I have not actually tested this either yet, only picking this up from code review)

(see https://doc.rust-lang.org/std/collections/hash_map/enum.Entry.html#method.or_insert)

DanNixon (changes requested):

trace-telemetry-exporter/src/main.rs
DanNixon: Could I ask you to test this before making it ready for review again.

Henry-Denny: Just from using a `println!` statement to print out the hashmap, it seems to be working correctly. Are there any other parts of the program that I should test?

DanNixon: I mean testing the entire program, i.e. run it, send some dummy trace data, `curl` the metrics endpoint and you should see metrics and see them change as new trace data arrives.

Henry-Denny: Thank you, it didn't occur to me that I could use `curl` to test this.

trace-telemetry-exporter/src/main.rs
```diff
@@ -0,0 +1,245 @@
...
+        recent_msg_counts
+            .iter()
+            .for_each(|(digitiser_id, recent_msg_count)| {
+                let msg_rate = *recent_msg_count as f64 / message_rate_interval as f64;
+                let labels = [get_digitiser_label(*digitiser_id)];
+                gauge!("digitiser_message_received_rate", &labels).set(msg_rate);
+            });
+
+        // Reset recent message count for each digitiser.
+        recent_msg_counts.clear();
```
DanNixon, line 150: Clearing this means that any previously reported digitiser remains on the last known message rate, which is a false reading if it is in fact not sending data.

Set values to zero instead to avoid this.

trace-telemetry-exporter/src/main.rs
```diff
@@ -0,0 +1,245 @@
...
+    time::Duration,
+};
+use supermusr_streaming_types::dat1_digitizer_analog_trace_v1_generated::{
+    digitizer_analog_trace_message_buffer_has_identifier, root_as_digitizer_analog_trace_message,
+    DigitizerAnalogTraceMessage,
+};
+use tokio::{task, time::sleep};
+use tracing::{debug, trace, warn};
+
+type MessageCounts = Arc<Mutex<HashMap<u8, usize>>>;
```
DanNixon, line 23: If the `u8` is the digitiser ID then make that explicit by using the type in the `common` crate.

DanNixon (changes requested):

events-to-histogram/src/main.rs
```diff
@@ -0,0 +1,154 @@
+mod metrics;
```
DanNixon, line 1: Why has this file been added?

DanNixon (changes requested):

events-to-histogram/src/main.rs
Henry-Denny: I think I must have messed up the merge. It's hard to tell what is no longer needed, I'll usually opt for keeping instead of discarding changes.

DanNixon (approved):

PR #115, opened 2024-02-22
Replace `log`/`env_logger` with `tracing`/`tracing_subscriber`

### LINKED ISSUE #99: Replace log/env_logger with tracing/tracing_subscriber

Henry-Denny:
Closes #99


common/src/lib.rs
```diff
@@ -19,6 +20,11 @@ pub struct EventData {
 
 pub const CHANNELS_PER_DIGITIZER: usize = 8;
 
+pub fn init_tracing_subscriber() {
+    let env_filter = EnvFilter::try_from_default_env().unwrap_or_else(|_| EnvFilter::new("info"));
```
DanNixon, line 24: You do not need this, `tracing_subscriber` will honour the level set by `RUST_LOG` by default.

DanNixon (changes requested):

common/src/lib.rs
Henry-Denny: This does not appear to work properly. The `env-filter` feature is included in the main `Cargo.toml` file but calling `tracing_subscriber::fmt().init()` at the beginning of `main` does not read from `RUST_LOG`.

This only seems to work if you use the full:

```rust
tracing_subscriber::fmt()
    .with_env_filter(EnvFilter::from_default_env())
    .init();
```

DanNixon: It does work, see https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/fn.init.html, the behaviour is dependant on the feature flags, without `env-filter` it respects `RUST_LOG` in the same way `env_logger` does.

Cargo.toml
```diff
@@ -47,4 +45,4 @@ supermusr-streaming-types = { path = "./streaming-types" }
 taos = { version = "0.10.27", default_features = false, features = ["ws"] }
 tokio = { version = "1.36", features = ["macros", "rt-multi-thread"] }
 tracing = "0.1.40"
-tracing-subscriber = "0.3.18"
+tracing-subscriber = { version = "0.3.18", features = ["env-filter"] }
```
DanNixon, line 48:
```suggestion
tracing-subscriber = "0.3.18"
```

digitiser-aggregator/src/main.rs
```diff
@@ -53,7 +53,7 @@ struct Cli {
 
 #[tokio::main]
 async fn main() {
-    tracing_subscriber::fmt::init();
+    tracing_subscriber::fmt().init();
```
DanNixon, line 56:
```suggestion
    tracing_subscriber::fmt::init();
```

events-to-histogram/src/main.rs
```diff
@@ -81,7 +81,7 @@ async fn main() -> Result<()> {
     loop {
         match consumer.recv().await {
             Ok(m) => {
-                log::debug!(
+                tracing::debug!(
```
DanNixon, line 84: For the sake of clarity, could the various macros be imported rather than using the fully qualified name every time.

See https://github.com/STFC-ICD-Research-and-Design/supermusr-data-pipeline/blob/main/digitiser-aggregator/src/main.rs#L18

DanNixon (changes requested):

kafka-daq-report/src/main.rs
```diff
@@ -101,7 +102,7 @@ enum Event<I> {
 #[tokio::main]
 async fn main() -> Result<()> {
     let args = Cli::parse();
-    log::debug!("Args: {:?}", args);
+    debug!("Args: {:?}", args);
```
DanNixon, line 105:
```suggestion
```

trace-archiver/src/main.rs
```diff
@@ -40,10 +41,10 @@ struct Cli {
 
 #[tokio::main]
 async fn main() -> Result<()> {
-    env_logger::Builder::from_env(env_logger::Env::default().default_filter_or("info")).init();
+    tracing_subscriber::fmt::init();
 
     let args = Cli::parse();
-    log::debug!("Args: {:?}", args);
+    debug!("Args: {:?}", args);
```
DanNixon, line 47:
```suggestion
```

DanNixon (approved):

PR #118, opened 2024-02-29
Trace to events Updates

Modularius:
A few bug fixes and refactoring.

Bug Fixes:
 - Added a function to specify the polarity of the trace (whether pulses register as positive or negative), and the baseline of the signal. This prevents the algorithms getting confused and registering an event at the first time-bin.
 - Refactored the process function arguments (and other local functions) to bundle a few related arguments into a struct.
 - Fixed a problem I introduced with the new baseline argument.
 - Refactored the constant phase discriminator command line arguments to get rid of the ThresholdDurationWrapper struct.
 - Added extra tests for the process function.
 - Fixed bug in the smoothing window. std::Vec extends its capacity in an unexpected way.

The tool has also been upgraded to include a few speed ups, including:
 - Streamline the process for writing channel event data to the message channel list
 - Rayon multithreading to process multiple channels simultaneously
 - Dispensed with pulse assembler in the case of constant phase discriminator (this did nothing extra)


trace-to-events/src/main.rs
```diff
@@ -38,7 +41,13 @@ struct Cli {
...
 
-    #[clap(long, default_value = "127.0.0.1:9090")]
+    #[clap(long)]
+    polarity: Polarity,
+
+    #[clap(long, default_value = "0")]
+    baseline: Intensity,
+
+    #[clap(long, env, default_value = "127.0.0.1:9090")]
     observability_address: SocketAddr,
```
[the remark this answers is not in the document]
DanNixon: Actually, why have the metrics been removed?

trace-to-events/src/main.rs
```diff
@@ -65,6 +70,7 @@ async fn main() {
     );
 
     let producer: FutureProducer = client_config
+        .set("linger.ms", "0")
```
DanNixon, line 73: Pretty sure the default value for `linger.ms` is 0ms anyway, is there a reason for setting this?

trace-to-events/src/pulse_detection/detectors/threshold_detector.rs
```diff
@@ -128,13 +106,13 @@ impl<Class: ThresholdClass> Assembler for ThresholdAssembler<Class> {
 
 #[cfg(test)]
 mod tests {
-    use super::*;
-    use crate::pulse_detection::EventFilter;
+    use super::{Data, ThresholdDetector, ThresholdDuration};
```
DanNixon, line 109:
```suggestion
    use super::*;
```

I'd prefer to keep this style throughout the codebase.

DanNixon (changes requested):
It is not so important for this PR as the work to correct it is probably more than justifiable, but in future for work where there are several smaller changes in one PR, please split them into atomic commits with messages that describe what has been done.

For now, could the specifics of the bug fixes be put into the PR body so at least they will be present in the squashed commit.

I also notice from the commit history you seem to have started this work from something other than `main`?

trace-to-events/src/main.rs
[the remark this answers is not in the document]
Modularius: I thought you said we were going to implement a different metrics solution in the future? I checked what you'd done on the events aggregator and saw you'd kept the observability address.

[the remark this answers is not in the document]
DanNixon: Ah, right, so I did, that was a mistake.

Could the metrics here either be left as they are, or immediately replaced with the `metrics` crate. Replacing them will be much easier with the existing implementation still in place.

trace-to-events/src/main.rs
```diff
@@ -50,13 +59,9 @@ struct Cli {
...
 #[tokio::main]
 async fn main() {
-    tracing_subscriber::fmt::init();
-
     let args = Cli::parse();
 
-    let mut watcher = Watcher::<AlwaysReady>::default();
-    metrics::register(&watcher);
-    watcher.start_server(args.observability_address).await;
+    //tracing_subscriber::fmt().init();
```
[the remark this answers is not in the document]
Modularius: I may have misunderstood but I thought we were using tracing for metrics now?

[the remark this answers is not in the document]
DanNixon: No, `tracing_subscriber` is what provides console logging currently.

trace-to-events/src/main.rs
```diff
@@ -38,9 +43,18 @@ struct Cli {
     #[clap(long)]
     event_topic: String,
 
-    #[clap(long, default_value = "127.0.0.1:9090")]
+    /// Determines whether events should register as positive or negative voltage
```
DanNixon, line 46: Could we avoid using the term "voltage" in favour of "intensity".

Any value in a trace is seen as an arbitrary intensity rather than a voltage given that no ADC scaling is applied.

trace-to-events/src/processing.rs
```diff
@@ -20,94 +23,108 @@ use supermusr_streaming_types::{
...
-                pulse.start.time.unwrap() as usize,
-                pulse.start.value.unwrap_or_default() as Intensity,
-            )
-        })
-        .collect()
+    let mut time = Vec::<Time>::new();
+    let mut voltage = Vec::<Intensity>::new();
+    for pulse in pulses {
+        time.push(pulse.0 as Time);
+        voltage.push(0 /*pulse.1. as Intensity*/);
```
DanNixon, line 104: Is this commented for a reason?

trace-to-events/src/processing.rs
```diff
@@ -180,42 +216,49 @@ fn get_save_file_name(path: &Path, channel: Channel, subscript: &str) -> PathBuf
...
+                            &trace.metadata(),
+                            &channel_trace,
+                            sample_time_in_ns,
+                            detector_settings,
+                            save_options,
+                        )
+                    }),
+                )
+            })
+            .collect();
```
DanNixon, line 253:
```suggestion
            .collect();

```

DanNixon (changes requested):

trace-to-events/src/main.rs
```diff
@@ -101,32 +115,43 @@ async fn main() {
...
-                                        Duration::from_secs(0),
                                     )
-                                    .await
-                                {
-                                    Ok(_) => {
-                                        trace!("Published event message");
-                                        metrics::MESSAGES_PROCESSED.inc();
+                                    .expect("Producer sends");
+
+                                tokio::spawn(async {
```
DanNixon, line 138: Does this actually help throughput? I would have expected the benefit of handling only the transmission of the event list message would not really be of much benefit when the detection and de/serialisation all happens sequentially.

To be clear, since event detection is already a trivially parallisable problem at the digitiser level, I am of the opinion that adding complexity to the event detector is not worth effort when it can be horizontally scaled with no changes required.

DanNixon (approved):

DanNixon:
@Modularius there are conflicts that need to be resolved here.

DanNixon (dismissed):

DanNixon:
@Modularius Is the description of this PR an accurate summary of what is in the PR?

Modularius:
Yes, I updated it to include all the bug fixes.

DanNixon:
I thought the multithreading had been removed?

Modularius:
No worries, is Rayon the crate with par_iter()? I haven't tried it, would it be better to use that?

DanNixon:
I think it would be better since it is less code to do the equivalent thing and is already used elsewhere in the project.

DanNixon (approved):

PR #122, opened 2024-03-07
Nexus writer

Modularius:
A functional tool which creates a NeXus file upon the start of a run, and appends event data to it when frame-aggregated or digitizer event list messages are created.


nexus-writer/diagrams/.gitignore
```diff
@@ -0,0 +1 @@
+!*.svg
```
DanNixon, line 1: Is there a need to have SVG in the root `.gitignore`?

DanNixon (changes requested):

nexus-writer/diagrams/.gitignore
Modularius: No, that folder is a duplicate of docs in fact

DanNixon: You will now need to alter the root `.gitignore`.

DanNixon (changes requested):
[see above]

PR #134, opened 2024-04-05
Open Telemetry Tracer Infrastructure 

Modularius:
Infrastructure to allow components to use Open Telemetry for tracing.

Tools should use
let _tracer = OtelTracer::new(end_point, service_name).expect("Tracer should be created");
in place of

This sets up the Open Telemetry subscriber, which is destroyed when _tracer goes out of scope. The destructor ensures any buffered traces are flushed.

The URI of the collector to send the traces is given by `end_point`, for instance `OtelTracer::new("http://localhost:4317/v1/traces", "MyService")`
N.B. "http://localhost:4317/v1/traces" is default endpoint for Jaeger.

A suitable collector, such as Jaeger, should be set up to receive the traces.

The generic type Spanned<T> is also provided to allow types which are associated with a particular span, to be created. For instance for the `Run` type in the NexusWriter crate, a trait is created
```
pub(crate) trait RunLike : Debug + AsRef<Run> + AsMut<Run> {
    fn new(filename: Option<&Path>, parameters: RunParameters, settings: &NexusSettings) -> Result<Self> where Self: Sized;
}
```
and is implemented for both `Run` and `Spanned<Run>`.
Any function or type that uses `Run`, instead uses a generic type `R : RunLike`, and calls `run.as_ref()` where `run : &R` to access the methods of run.
This way Run and Spanned<Run> can be used interchangeably with no overhead cost.


common/src/tracer.rs
```diff
@@ -0,0 +1,158 @@
...
+
+const ENDPOINT: &str = "http://localhost:4317/v1/traces";
+
+/// Create this object to initialise the Open Telemetry Tracer
+pub struct OtelTracer;
+
+impl OtelTracer {
+    pub fn new(service_name: &str) -> Result<Self, TraceError> {
+        Self::new_with_endpoint(ENDPOINT, service_name)
+    }
```
DanNixon, line 20:
```suggestion
    }

```

Cargo.toml
```diff
@@ -47,4 +50,6 @@ taos = { version = "0.10.27", default_features = false, features = ["ws"] }
 tokio = { version = "1.37", features = ["macros", "rt-multi-thread"] }
 thiserror = "1.0.58"
 tracing = "0.1.40"
+tracing-opentelemetry = "0.23.0"
 tracing-subscriber = "0.3.18"
+
```
DanNixon, line 55:
```suggestion
```

common/src/tracer.rs
```diff
@@ -0,0 +1,158 @@
...
+    trace::{TraceContextExt, TraceError},
+};
+use opentelemetry_otlp::WithExportConfig;
+use rdkafka::message::{BorrowedHeaders, Headers, OwnedHeaders};
+use std::fmt::Debug;
+use tracing::Span;
+use tracing_opentelemetry::{self, OpenTelemetrySpanExt};
+use tracing_subscriber::layer::SubscriberExt;
+
+const ENDPOINT: &str = "http://localhost:4317/v1/traces";
```
DanNixon, line 12: I would rather this default be handled in command line parsing of the tool that uses OpenTelemetry, this makes it more explicit on the CLI and reduces boilerplate code.

common/src/tracer.rs
```diff
@@ -0,0 +1,158 @@
...
+    }
+}
+
+impl Drop for OtelTracer {
+    fn drop(&mut self) {
+        opentelemetry::global::shutdown_tracer_provider()
+    }
+}
+
+pub struct Spanned<T> {
```
DanNixon, line 76: What is this used for?

DanNixon (changes requested):

Cargo.toml
DanNixon: Could the extra blank line be removed.

common/src/tracer.rs
DanNixon: Could that be put into some struct documentation.

DanNixon (changes requested):

common/src/tracer.rs
DanNixon: @Modularius

DanNixon (approved):

PR #135, opened 2024-04-05
Update simulator

Modularius:
A major update to the simulator tool which allows complex traces to be simulated and dispatched to the Kafka broker. The update is non-breaking, i.e. "single" and "continuous" mode can still be used and are unchanged.
The settings are read from a json file, an example of which is:
```json
{
    "voltage-transformation": {"scale": 1, "translate": 0 },
    "traces": [
        {
            "digitizers": [ { "id": 0, "channels": { "min": 0, "max": 1 } }],
            "frames": { "vector": [1, 2, 3, 4, 5] },
            "sample-rate": 100000000,
            "pulses": [
                {
                    "weight": 1,
                    "attributes": {
                        "type": "biexp",
                        "height": { "type": "uniform", "min": { "fixed": 30 }, "max": { "fixed": 70 } },
                        "start":  { "type": "exponential", "lifetime": { "fixed": 2200 } },
                        "rise":   { "type": "uniform", "min": { "fixed": 20 }, "max": { "fixed": 30 } },
                        "decay":  { "type": "uniform", "min": { "fixed": 5 }, "max": { "fixed": 10 } }
                    }
                },
                {
                    "weight": 1,
                    "attributes": {
                        "type": "flat",
                        "start":  { "type": "exponential", "lifetime": { "fixed": 2200 } },
                        "width":  { "type": "uniform", "min": { "fixed": 20 }, "max": { "fixed": 50 } },
                        "height": { "type": "uniform", "min": { "fixed": 30 }, "max": { "fixed": 70 } }
                    }
                },
                {
                    "weight": 1,
                    "attributes": {
                        "type": "triangular",
                        "start":     { "type": "exponential", "lifetime": { "fixed": 2200 } },
                        "width":     { "type": "uniform", "min": { "fixed": 20 }, "max": { "fixed": 50 } },
                        "peak_time": { "type": "uniform", "min": { "fixed": 0.25 }, "max": { "fixed": 0.75 } },
                        "height":    { "type": "uniform", "min": { "fixed": 30 }, "max": { "fixed": 70 } }
                    }
                }
            ],
            "noises": [
                {
                    "attributes": { "type" : "gaussian", "mean" : { "fixed": 0 }, "sd" : { "fixed": 20 } },
                    "smoothing-factor" : { "fixed": 0.975 },
                    "bounds" : { "min": 0, "max": 30000 }
                },
                {
                    "attributes": { "type" : "gaussian", "mean" : { "fixed": 0 }, "sd" : { "frame-transform": { "scale": 50, "translate": 50 } } },
                    "smoothing-factor" : { "fixed": 0.995 },
                    "bounds" : { "min": 0, "max": 30000 }
                }
            ],
            "num-pulses": { "type": "constant", "value": { "fixed": 500 } },
            "time-bins": 30000,
            "timestamp": "now",
            "frame-delay-us": 20000
        }
    ]
}
```


simulator/README.md
```diff
@@ -0,0 +1,161 @@
...
+
+```shell
+simulator --help
+```
+
+### Commands
+
+- `Single`:            Produce a single trace.
+- `Continuous`:        Produce a regular infinite sequence of traces are regular intervals.
+- `Json`:              Produce traces in the manner specified by a json file.
```
DanNixon, line 32:
```suggestion
- `single`:            Produce a single trace.
- `continuous`:        Produce a regular infinite sequence of traces are regular intervals.
- `json`:              Produce traces in the manner specified by a json file.
```

simulator/README.md
```diff
@@ -0,0 +1,161 @@
...
+
+## Json Format
```
DanNixon, line 34: Can we call this something other than the format the data is defined in, e.g. "arbitrary", "defined", or similar.

simulator/README.md
```diff
@@ -0,0 +1,161 @@
...
+
+In `json` mode, the behavior is given by the simulator object in the user-defined json file.
+The structure of the simulator object is:
+
+### Simulator
+
+- voltage: `Interval`
```
DanNixon, line 41: Could you add relative links to these types to make navigating this on the GitHub render easier.

simulator/src/json.rs
```diff
@@ -0,0 +1,220 @@
...
+            Expression::Fixed(v) => *v,
+            Expression::FrameTransform(frame_function) => {
+                frame_function.transform(frame_index as f64)
+            }
+        }
+    }
+}
+
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "kebab-case", untagged)]
```
DanNixon, line 27: I'm quite against `untagged` as it adds a layer of obfuscation as to what you are defined, prefer `tag = "type"` as you have used elsewhere.

simulator/src/main.rs
```diff
@@ -1,10 +1,21 @@
+mod json;
```
DanNixon, line 1: Similar as above, rename this to what it is, not the format it is defined in.

simulator/src/main.rs
```diff
@@ -127,6 +159,70 @@ async fn main() {
                 frame.tick().await;
             }
         }
+        Mode::Json(Json { path, repeat }) => {
```
DanNixon, line 162: This block is large enough to be it's own function. At which point the other two match arms may as well be split out too.

simulator/src/muon.rs
```diff
@@ -0,0 +1,210 @@
+use super::json;
+use supermusr_common::{Intensity, Time};
+
+#[derive(Debug)]
+pub(crate) enum Muon {
```
DanNixon, line 5: This name is a tad ambiguous, I assume this is modelling a muon decay?

DanNixon (changes requested):

simulator/src/muon.rs
Modularius: This models a muon detection event.

simulator/src/json.rs
Modularius: I used `untagged`, as fields of type `Expression`, will automatically coerce a float to Expression::Fixed(float). This allows the user to write either a literal or a `Transformation` (a linear function depending on the frame number). As most values are likely to be literals this avoids the user having to clutter up the file with things like `{"type": "Fixed", "value": 35.0}`.
The same applies to fields of type `RandomDistribution` and `Frames`, though to a lesser extent, so perhaps it makes sense to only use `untagged` for `Expression`?

DanNixon: I don't think the benefits of such syntactic niceties are worth the validation/type safety that explicit specifying what you want bring. I would rather not use `untagged` anywhere.

simulator/src/muon.rs
DanNixon: In that case could this be renamed to something that conveys that.

simulator/src/defined.rs
```diff
@@ -0,0 +1,308 @@
...
+                }
+            }
+        } else {
+            f64::default()
+        }
+    }
+}
+
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "kebab-case", tag = "noise-type")]
```
DanNixon, line 129: Does the prefix gain anything?

simulator/src/main.rs
```diff
@@ -87,6 +108,17 @@ struct Continuous {
     frame_time: u64,
 }
 
+#[derive(Clone, Parser)]
+struct Defined {
+    /// Path to the json settings file
+    #[clap(long)]
+    path: PathBuf,
```
DanNixon, line 115:
```suggestion
    file: PathBuf,
```

Also this would be better as an argument rather than an option.

simulator/src/main.rs
```diff
@@ -1,10 +1,21 @@
+mod defined;
```
DanNixon, line 1: This is not a descriptive name, something like `simulation_config` would be much better.

DanNixon (changes requested):

simulator/src/defined.rs
```diff
@@ -0,0 +1,308 @@
...
+            Expression::FixedValue(v) => *v,
+            Expression::FrameTransform(frame_function) => {
+                frame_function.transform(frame_index as f64)
+            }
+        }
+    }
+}
+
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "kebab-case", tag = "random-type")]
```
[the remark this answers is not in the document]
Modularius: It's more descriptive, so its easier to understand the config file when you are reading/editing it.

simulator/src/main.rs
Modularius: I'm not sure what you mean, as this is not an option. Do you mean it should be in the main Cli struct? It's only used for this mode however.

simulator/src/main.rs
```diff
@@ -44,6 +55,13 @@ struct Cli {
     #[clap(long)]
     trace_topic: Option<String>,
 
+    #[clap(long, env, default_value = "127.0.0.1:9090")]
+    observability_address: SocketAddr,
+
+    /// Topic to publish analog trace packets to
```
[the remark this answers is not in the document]
Modularius: This relic code, and has now been removed.

simulator/src/main.rs
DanNixon: This would commonly be regarded as an option (claps autogenerated help even does so) as it requires an identifier.

e.g. in `ls --sort size Downloads`, `--sort size` is an option and `Downloads` is an argument.

If you always need to specify something and it is the main thing you are acting on, then it is better to specify it as an argument.

Modularius: Ah, I see now

simulator/src/main.rs
```diff
@@ -87,7 +83,7 @@ enum Mode {
     Continuous(Continuous),
 
     /// Run in json mode, behaviour is defined by the file given by --path
-    Defined(Defined),
+    SimulationConfig(SimulationConfig),
```
DanNixon, line 86: This was fine as it was, as this is for the CLI. The change was specific to the name of the `simulation_config` module.

DanNixon (changes requested):

Cargo.toml
```diff
@@ -24,8 +24,9 @@ edition = "2021"
 [workspace.dependencies]
 anyhow = "1.0"
 assert_approx_eq = "1.1.0"
-async-trait = "0.1.79"
-chrono = {version = "0.4.37", features = ["serde"] }
+
```
DanNixon, line 27:
```suggestion
```

Cargo.toml
```diff
@@ -24,8 +24,9 @@ edition = "2021"
...
+async-trait = "0.1.80"
+chrono = {version = "0.4.38", features = ["serde"] }
```
DanNixon, line 29:
```suggestion
chrono = { version = "0.4.38", features = ["serde"] }
```

DanNixon (changes requested):

simulator/src/main.rs
```diff
@@ -87,6 +100,17 @@ struct Continuous {
     frame_time: u64,
 }
 
+#[derive(Clone, Parser)]
+struct Defined {
+    /// Path to the json settings file
+    #[clap()]
```
DanNixon, line 106:
```suggestion
```

simulator/src/simulation_config.rs
```diff
@@ -0,0 +1,308 @@
...
+            Expression::FixedValue(v) => *v,
+            Expression::FrameTransform(frame_function) => {
+                frame_function.transform(frame_index as f64)
+            }
+        }
+    }
+}
+
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "kebab-case", tag = "random-type")]
```
DanNixon, line 27: Do the prefixes on the tag string really add anything or can they all just be "type"?

DanNixon (changes requested):

simulator/src/simulation_config.rs
Modularius: I believe so, they make it easier to use the configuration file.

DanNixon (approved):

PR #148, opened 2024-04-22
Add open telemetry to digitiser aggregator

Modularius:
Adds OpenTelemetry and some tracing to the aggregator component.

- otel_endpoint optional command line parameter added
- init_tracer function sets up OpenTelemetry if otel_endpoint is specified and runs tracing_subscriber::fmt::init() otherwise
- Extracts external trace from consumed Kafka headers if present and sets current trace as a child of it.
- Associates a trace with each Frame object, to which all relevant external traces are collected. (I'm not 100% happy with how fiddly this is - modifications are needed to the PartialFrame and AggregatedFrame functionality - it suggests there could be room to streamline the Spanned generic type in the future, but it works fine for the moment).

### DIFF
--- a/digitiser-aggregator/src/main.rs
+++ b/digitiser-aggregator/src/main.rs
@@ -111,6 +111,7 @@ async fn main() {
     }
 }
 
+#[tracing::instrument(skip_all, name = "Event Formation Message", level = "trace")]
 async fn on_message(
     root_span: &Span,
     args: &Cli,
@@ -118,9 +119,6 @@ async fn on_message(
     producer: &FutureProducer,
     msg: &BorrowedMessage<'_>,
 ) {
-    let span = trace_span!("Event Formation Message");
-    let _guard = span.enter();
-
     if args.otel_endpoint.is_some() {
         if let Some(headers) = msg.headers() {
             debug!("Kafka Header Found");
--- a/digitiser-aggregator/src/main.rs
+++ b/digitiser-aggregator/src/main.rs
@@ -133,14 +133,13 @@ async fn on_message(
             match root_as_digitizer_event_list_message(payload) {
                 Ok(msg) => {
                     debug!("Event packet: metadata: {:?}", msg.metadata());
-                    root_span.in_scope(|| {
-                        cache.push(
-                            msg.digitizer_id(),
-                            msg.metadata().try_into().unwrap(),
-                            msg.into(),
-                        );
-                    });
+                    cache.push(
+                        msg.digitizer_id(),
+                        msg.metadata().try_into().unwrap(),
+                        msg.into(),
+                    );
                     if let Some(frame) = cache.find(msg.metadata().try_into().unwrap()) {
+                        OtelTracer::set_span_parent_to(&frame.span, root_span);
                         let cur_span = tracing::Span::current();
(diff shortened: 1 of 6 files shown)


digitiser-aggregator/src/main.rs
```diff
@@ -83,33 +90,44 @@ async fn main() {
...
 
 async fn on_message(
+    root_span: &Span,
+    args: &Cli,
     cache: &mut FrameCache<EventData>,
     producer: &FutureProducer,
-    output_topic: &str,
     msg: &BorrowedMessage<'_>,
 ) {
+    let span = trace_span!("Event Formation Message");
```
DanNixon, line 121: This is a case that should just be done with a function annotation, not manually creating a span.

digitiser-aggregator/src/spanned_frame.rs
```diff
@@ -0,0 +1,64 @@
...
+    fn new(ttl: Duration, metadata: FrameMetadata) -> Self {
+        let span = trace_span!("Frame");
+        Spanned {
+            span,
+            value: PartialFrame::<D>::new(ttl, metadata),
+        }
+    }
+}
+
+pub(crate) struct SpannedAggregatedFrame<D> {
```
DanNixon, line 23: Why can't `AggregatedFrame` also use `Spanned`?

DanNixon (changes requested):

digitiser-aggregator/src/spanned_frame.rs
Modularius: Due to Rust rules on implementing foreign traits on foreign types. I'm going to redesign this in a subsequent PR however.

DanNixon: I don't see any reason not to have the trait impl next to the type definition in this case though. The defining crate is in this repo and the fact there is an additional trait on the type should not adversely affect any other crate.

Modularius: It's a restriction on foreign types *and* foreign traits, so as Spanned<T> is from another crate, it won't allow compilation.
That is
```rust
SpannedAggregatedFrame<D> = Spanned<AggregatedFrame<D>>;

impl<D: Debug> AggregatedFrameLike<D> for Spanned<AggregatedFrame<D>> {
...
}
```
is not allowed.

DanNixon: Ah, I see, it's the `impl From` that is the issue. If this is going to be redesigned then it is fine for now.

digitiser-aggregator/src/main.rs
```diff
@@ -83,33 +90,42 @@ async fn main() {
...
             }
             _ = cache_poll_interval.tick() => {
-                cache_poll(&mut cache, &producer, &args.output_topic).await;
+                cache_poll(&args, &mut cache, &producer).await;
             }
         }
     }
 }
 
+#[tracing::instrument(skip_all, name = "Event Formation Message", level = "trace")]
```
DanNixon, line 114:
```suggestion
#[tracing::instrument(skip_all, level = "trace")]
```

I'd prefer just sticking with function names, if that is not clear when looking at the traces then the function should be renamed.

PR #150, opened 2024-04-22
Add open telemetry to simulator

Modularius:
Adds OpenTelemetry and some tracing to the trace simulator component.

- otel_endpoint optional command line parameter added
- init_tracer function sets up OpenTelemetry if otel_endpoint is specified and runs tracing_subscriber::fmt::init() otherwise
- Propagates current trace to the next component via the Kafka header.

### DIFF
--- a/simulator/src/main.rs
+++ b/simulator/src/main.rs
@@ -212,12 +212,12 @@ async fn send(
                     &mut headers,
                 );
 
-                FutureRecord::to(&topic)
+                FutureRecord::to(topic)
                     .payload(fbb.finished_data())
                     .headers(headers)
                     .key("Simulated Event")
             } else {
-                FutureRecord::to(&topic)
+                FutureRecord::to(topic)
                     .payload(fbb.finished_data())
                     .key("Simulated Event")
             }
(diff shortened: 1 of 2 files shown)


simulator/src/main.rs
```diff
@@ -415,3 +464,19 @@ async fn run_configured_simulation(cli: &Cli, producer: &FutureProducer, defined
         }
     }
 }
+
+fn init_tracer(otel_endpoint: Option<&str>) -> Option<OtelTracer> {
```
DanNixon, line 468: This can be made sufficiently generic to go into `common` with only adding a couple of parameters.

In fact, if you made it a macro you could get away with the application name being taken from the crate name.

simulator/src/message.rs
```diff
@@ -38,7 +39,6 @@ impl<'a> TraceMessage {
         .attributes
     }
 
-    #[tracing::instrument]
```
DanNixon, line 41: Why have these been removed? Assuming they create spans (which was my understanding) it seems like a low effort way to get more fine grained information.

As a side note, assuming my assumptions are correct, I would prefer if this was used as the main way to instrument the code for tracing going forward given that it promotes appropriately sized, well organised functions.

DanNixon (approved):

PR #155, opened 2024-04-25
Remove histogram functionality

DanNixon:
This is now going to be handled in tools downstream of the NeXus file, as such is out of scope for the data pipeline itself.


Modularius (approved):
Tested successfully with Jaeger, no change in behaviour.

PR #157, opened 2024-04-25
More generic tracer code

Modularius:
- Individual `init_trace` changed to `conditional_init_trace` macro in common crate. This sets up open telemetry if given a `Some(otel_endpoint)` parameter, and uses the default trace subscriber if given `None`.
 - Code for propagating OpenTelemetry traces via Kafka headers is simplified by folding the functions into traits FutureRecordTracerExt and OptionalHeaderTracerExt, used for producing and consuming Kafka messages respectively.
 - Unused function removed from tracer.rs.
 - Tested using Jaeger UI with no change in behavior.


common/src/tracer.rs
```diff
@@ -1,14 +1,101 @@
...
+    fn conditional_extract_to_current_span(self, use_otel: bool);
+    fn conditional_extract_to_span(self, use_otel: bool, span: &Span);
+}
+
+impl OptionalHeaderTracerExt for Option<&BorrowedHeaders> {
+    fn conditional_extract_to_current_span(self, use_otel: bool) {
+        self.conditional_extract_to_span(use_otel, &tracing::Span::current())
+    }
+
+    fn conditional_extract_to_span(self, use_otel: bool, span: &Span) {
```
DanNixon, line 87: What does this do if `use_otel` is `true`, the message has headers but not headers that were added by OpenTelemetry?

digitiser-aggregator/src/main.rs
```diff
@@ -119,12 +127,8 @@ async fn on_message(
...
     msg: &BorrowedMessage<'_>,
 ) {
-    if args.otel_endpoint.is_some() {
-        if let Some(headers) = msg.headers() {
-            debug!("Kafka Header Found");
-            OtelTracer::extract_context_from_kafka_to_span(headers, &tracing::Span::current());
-        }
-    }
+    msg.headers()
+        .conditional_extract_to_current_span(args.otel_endpoint.is_some());
```
DanNixon, line 131: `_tracer.is_some()` is potentially a clearer test to use here (and in similar checks elsewhere).

common/src/tracer.rs
Modularius: If it can't find valid OpenTelemetry headers, then it sets the parent of `span` to the current span, effectively it does nothing in this case.

Modularius: I will make a modification however, so that it adds to existing headers, rather than overwriting them. However it must push them to the front of the headers as this is where OpenTelemetry reads them from.

common/src/tracer.rs
```diff
@@ -1,14 +1,101 @@
...
+use tracing::{debug, level_filters::LevelFilter, Span};
 use tracing_opentelemetry::OpenTelemetrySpanExt;
 use tracing_subscriber::{filter, layer::SubscriberExt, Layer};
 
+/// Should be called at the start of each component
+/// The `conditional_` prefix used in the methods of FutureRecordTracerExt and OptionalHeaderTracerExt
+/// indicate the method's first parameter is a bool, however here the first parameter is an Option<&str>
+/// with the URL of the OpenTelemetry collector to be used, or None, if OpenTelemetry is not used.
+#[macro_export]
+macro_rules! conditional_init_tracer {
```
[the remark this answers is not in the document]
Modularius: Ah, module_path!() uses hyphens, whilst env!("CARGO_BIN_NAME") uses underscores. I'll use env!("CARGO_BIN_NAME") for the service_name, it's one less point of failure.

DanNixon (approved):

PR #160, opened 2024-04-26
Simplify spanned implementation

Modularius:
- Changed the `Spanned` generic type to the `SpanWrapped` generic type.
- Replaced `RunLike`, [...]`FrameLike` traits with the much simpler `Spanned` and `SpannedMut` traits.
- Introduced the `SpanOnce` type which can contain a span that can be attached to a type implementing Spanned. SpanOnce is designed to prevent misuse.
- Moved relevant code from tracer.rs to spanned.rs
- Created tests to ensure `SpanOnce` behaves as expected.

If you wish to attach a span to your type, you can:
- use the SpanWrapped type, or
- add a `span: SpanOnce` field, and implement the Spanned/SpannedMut traits.
If you do the latter, then only implement SpannedMut if you need to move the attached span, to a different type, as is the case with PartialFrame and AggregateFrame.
Spanned/SpannedMut allows one to refer to the `SpanOnce` field, which can be read, moved (taken) once, or initialised once.


digitiser-aggregator/src/frame/aggregated.rs
```diff
@@ -1,45 +1,45 @@
...
+
 impl<D> From<PartialFrame<D>> for AggregatedFrame<D>
 where
     DigitiserData<D>: Accumulate<D>,
 {
     fn from(mut partial: PartialFrame<D>) -> Self {
         Self {
+            span: partial
+                .span_mut()
+                .take()
```
DanNixon, line 41: If I am reading `spanned.rs` correctly then this should only give an error in the event of a logic error in the code, not some unknown external factor, correct?

In which case simply using `unwrap()` is fine.

At some point in the future it might be worth revisiting the error handing options, which I'll describe in https://github.com/STFC-ICD-Research-and-Design/supermusr-data-pipeline/issues/161

Modularius: Yes, it's just meant to enforce proper use of the SpanOnce type.

DanNixon (approved):
:+1:

PR #165, opened 2024-05-01
Remove handling of events from `stream-to-file`.

Henry-Denny:
**Overview of Changes:**

- `stream-to-file` has been renamed to `trace-archiver-hdf5`.
- The handling of events functionality has been removed from this tool.

This partially resolves #113.

PR #167, opened 2024-05-02
Trace Saving Solution.

### LINKED ISSUE #113: Trace saving solution
The existing [`stream-to-file`](https://github.com/STFC-ICD-Research-and-Design/supermusr-data-pipeline/tree/main/stream-to-file/src) tool should be adapted to handle saving traces for development/diagnostics.

- Remove handling of events, this is handled by the dedicated NeXus writer
- Allow the ability to consume a control topic, this will handle receiving run start and run stop messages
- If no control topic is provided then the tool behaves as it does now, i.e. continuously save data until it is terminated
- If a control topic is provided then the tool outputs a new HDF file for each run (i.e. from when a run start message is received, up until a run stop message is received)

Henry-Denny:
This work is a continuation of #165 and aims to close #113. It gives the `trace-archiver-hdf5` the ability to consume a control topic, and use this to handle when to start and stop the recording of trace data to an HDF file.


trace-archiver-hdf5/src/main.rs
```diff
@@ -121,7 +123,28 @@ async fn main() -> Result<()> {
                                     .inc();
                             }
                         }
+                    } else if args.control_topic == Some(msg.topic().to_string()) {
+                        // A message has been received from the control topic.
+                        if run_start_buffer_has_identifier(payload) {
+                            debug!("New run start.");
+                            // Start recording trace data to file.
+                            if file.is_none() {
+                                let filename = PathBuf::from(
```
DanNixon, line 132: Is the filename just the timestamp in milliseconds? Could something more human readable be used instead (ideally [RFC 3339](https://datatracker.ietf.org/doc/html/rfc3339)).

DanNixon (changes requested):

trace-archiver-hdf5/README.md
```diff
@@ -0,0 +1,8 @@
+# trace-archiver-hdf5
+
+This tool listens on a given trace topic and saves each received message inside an HDF5 file.
+If *no* control topic is provided,
+messages are continuously saved to the file specified by the `--file` command line argument,
+until the program terminates.
+If a control topic is provided,
+a new HDF5 file is created for each run with the file name `<timestamp>.hdf5`.
```
DanNixon, line 8:
```suggestion
a new HDF5 file is created for each run with the file name `<timestamp>.h5`.
```

trace-archiver-hdf5/src/main.rs
```diff
@@ -71,9 +81,22 @@ async fn main() -> Result<()> {
...
+    .collect::<Vec<&str>>();
 
-    let mut trace_file = TraceFile::create(&args.file, args.digitizer_count)?;
+    consumer.subscribe(&topics_to_subscribe)?;
+
+    let mut file = match args.control_topic {
+        // If a control topic is provided, a file will be created for each run.
+        Some(_) => None,
+        // If a a control topic is not provided, a persistent file is used instead.
+        None => Some(TraceFile::create(&args.file, args.digitizer_count)?),
```
DanNixon, line 98: What does `args.file` do when a control topic is provided?

trace-archiver-hdf5/src/main.rs
```diff
@@ -136,3 +158,35 @@ async fn main() -> Result<()> {
         };
     }
 }
+
+fn generate_filename(timestamp: Timestamp) -> PathBuf {
+    //  TODO: Check this unwrap does not cause any issues.
+    let timestamp =
+        DateTime::<Utc>::from_timestamp_millis(timestamp.to_millis().unwrap_or_default())
+            .unwrap_or_default();
+    PathBuf::from(format!("{:?}.hdf5", timestamp))
```
DanNixon, line 167:
```suggestion
    PathBuf::from(format!("{:?}.h5", timestamp))
```

trace-archiver-hdf5/src/main.rs
```diff
@@ -136,3 +158,35 @@ async fn main() -> Result<()> {
         };
     }
 }
+
+fn generate_filename(timestamp: Timestamp) -> PathBuf {
+    //  TODO: Check this unwrap does not cause any issues.
+    let timestamp =
```
DanNixon, line 164: Could something more sensible be done when unwrapping here?

trace-archiver-hdf5/src/main.rs
```diff
@@ -121,7 +125,25 @@ async fn main() -> Result<()> {
                                     .inc();
                             }
                         }
+                    } else if args.control_topic == Some(msg.topic().to_string()) {
+                        // A message has been received from the control topic.
+                        if run_start_buffer_has_identifier(payload) {
+                            debug!("New run start.");
+                            // Start recording trace data to file.
+                            if file.is_none() {
+                                let filename = generate_filename(msg.timestamp());
```
DanNixon, line 134: The timestamp from the run start message should be used here, not the Kafka timestamp.

DanNixon (changes requested):

trace-archiver-hdf5/run.hdf5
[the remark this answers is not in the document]
Henry-Denny: No sorry, that must have slipped through.

trace-archiver-hdf5/src/main.rs
Henry-Denny: Currently nothing, which is a problem since it is required regardless of what "mode" the tool is running in. I am trying to find a nice way to only make this required if a control topic is not provided.

Henry-Denny:
# Current state of PR

The work done in this pull request currently satisfies all of the requirements of #113. The only work to be done is the current outstanding requested [change](https://github.com/STFC-ICD-Research-and-Design/supermusr-data-pipeline/pull/167#discussion_r1629487562), that is, using the run start timestamp rather than the Kafka timestamp for the generated file.

Issue #161, opened 2024-04-26
Reconsider error handling options


DanNixon:
Currently this is loosely followed:

1. This should never happen under normal or abnormal execution: `unwrap()`
2. This might fail, we cannot recover and don't care about any state (usually in setup procedures): `expect()`
3. This might fail and the callee needs to care (i.e. in a library): `thiserror`
4. This might fail and the user might care (i.e. in a binary): `anyhow`

Ideally, the following changes should be made:

- Move cases of 2 into 4
- Move cases of 1 into 2, where the panic message explains why the condition should never happen
- Forbid `unwrap()`

DanNixon:
https://github.com/yassineldeeb/rust-panic-analyzer might be a useful tool to adopt to assist in enforcing this.

DanNixon:
The actual reconsideration has been done and documentation updated as appropriate. Smaller PRs will be created to address the migration of existing code.

PR #168, opened 2024-05-02
Fixes to Aggregator and Nexus Writer

Modularius:
- Spans encompassing Runs and Frames were not initialised when they should have been. This has now been fixed by changes to aggregator and nexus-writer and spanned.rs.
- ~~Code added to "tracer.rs" and components (trace-to-events, aggregator, simulator) to gracefully shut down the tracer should Err be sent to expect (e.g. in the event of a broker failure).~~
- This code will be added to nexus-writer, run-simulator and trace-reader in a later PR.

### DIFF
--- a/common/src/tracer.rs
+++ b/common/src/tracer.rs
@@ -1,5 +1,3 @@
-use std::fmt::{Debug, Display};
-
 use opentelemetry::{
     propagation::{Extractor, Injector},
     trace::TraceError,
@@ -9,7 +7,7 @@ use rdkafka::{
     message::{BorrowedHeaders, Headers, OwnedHeaders},
     producer::FutureRecord,
 };
-use tracing::{debug, error, level_filters::LevelFilter, Span};
+use tracing::{debug, level_filters::LevelFilter, Span};
 use tracing_opentelemetry::OpenTelemetrySpanExt;
 use tracing_subscriber::{filter, layer::SubscriberExt, Layer};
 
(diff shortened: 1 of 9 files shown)


common/src/tracer.rs
```diff
@@ -159,6 +161,12 @@ impl OtelTracer {
     pub fn set_span_parent_to(span: &Span, parent_span: &Span) {
         span.set_parent(parent_span.context());
     }
+
+    pub fn kill_tracer_on_err(e: impl Debug + Display) -> impl Debug + Display {
```
DanNixon, line 165: This feels quite clunky and I am not a fan of the way this function is used in error handling.

Why is it actually needed? Is it a result of the fact panics are used for some error handing in initialisation?

Modularius: Yes, if the program panics then spans or error/tracing messages may not be emitted. This seemed the most elegant solution, an alternative however would be handling those initialisation panics in code.

DanNixon: If it is specifically panicking that is the issue then I would prefer not doing this in favour of https://github.com/STFC-ICD-Research-and-Design/supermusr-data-pipeline/issues/161 (where the only reason for a panic would be a programming error that would realistically be an issue anyway).

Modularius: Ah of course, that would cause the tracer to shut down, eliminating the need for `OtelTracer::kill_tracer_on_err` would it not?

PR #172, opened 2024-05-10
Update nexus writer

Modularius:
Update to Nexus Writer
- Added runlog data receiving
- Added selog and alert message receiving
- ~~Added OpenTelemetry graceful shutdown on error~~

Should be merged after "Fixes to Aggregator and Nexus Writer" PR


DanNixon:
You seem to have both code for `nexus-writer` and `run-simulator` in here?

nexus-writer/src/main.rs
```diff
@@ -176,86 +155,172 @@ async fn main() -> Result<()> {
...
+        match $run.span().get() {
+            Ok(run_span) => run_span.in_scope(|| {
+                trace_span!($span_name).follows_from(cur_span);
+            }),
+            Err(e) => debug!("No run found. Error: {e}"),
+        }
+    };
+}
+
+impl NexusEngine {
```
DanNixon, line 171: Why is this `impl` here and not in `engine.rs`?

nexus-writer/src/nexus/hdf5_file/mod.rs
```diff
@@ -1,8 +1,9 @@
-mod event_run_file;
 mod hdf5_writer;
 mod run_file;
+mod run_file_components;
+
+use run_file_components::EventRun;
```
DanNixon, line 6:
```suggestion
```

nexus-writer/src/nexus/hdf5_file/run_file_components/runlog_file.rs
```diff
@@ -0,0 +1,61 @@
...
+use crate::nexus::{
+    hdf5_file::{
+        hdf5_writer::create_resizable_dataset,
+        run_file_components::timeseries_file::get_dataset_builder,
+    },
+    nexus_class as NX,
+};
+use anyhow::Result;
+use hdf5::{Group, SimpleExtents};
+use std::fmt::Debug;
```
DanNixon, line 11:
```suggestion
```

This is `std::prelude` so you do not need to explicitly `use` it.

nexus-writer/src/nexus/hdf5_file/run_file_components/selog_file.rs
```diff
@@ -0,0 +1,116 @@
...
+    hdf5_file::{
+        hdf5_writer::create_resizable_dataset,
+        run_file_components::timeseries_file::get_dataset_builder,
+    },
+    nexus_class as NX,
+};
+use anyhow::Result;
+use hdf5::{types::VarLenUnicode, Group, SimpleExtents};
+use ndarray::s;
+use std::fmt::Debug;
```
DanNixon, line 12:
```suggestion
```

nexus-writer/src/nexus/hdf5_file/run_file_components/timeseries_file.rs
```diff
@@ -0,0 +1,294 @@
+use anyhow::{anyhow, bail, Result};
+use chrono::{DateTime, Utc};
+use hdf5::{
+    types::{FloatSize, IntSize, TypeDescriptor},
+    Dataset, DatasetBuilderEmpty, Group, H5Type,
+};
+use ndarray::{s, Dim, SliceInfo, SliceInfoElem};
+use std::fmt::Debug;
```
DanNixon, line 8:
```suggestion
```

DanNixon (changes requested):

nexus-writer/src/main.rs
Modularius: This impl extends the impl block in engine.rs, as it creates the spans which collects different FrameEventLists into a single trace, and it handles the context propagation using functions from OtelTracer. I've tried to keep references to OtelTracer exclusively in the main.rs file wherever possible.

DanNixon: > I've tried to keep references to OtelTracer exclusively in the main.rs file wherever possible.

I'm not sure this really does much for readability over having `impl` blocks near the `struct` they refer to.

Moving it would also keep the size of `main.rs` down, right now it is starting to get large.

nexus-writer/src/nexus/hdf5_file/run_file_components/timeseries_file.rs
Modularius: It's needed here as it's explicitly used in line 20.
```rust
pub(super) trait TimeSeriesDataSource<'a> : Debug {
```

nexus-writer/src/main.rs
Modularius: The engine.rs file is even larger. I think it makes most sense to convert these from methods to functions and leave them in main.rs. We can probably compact it at a later date using macros.

DanNixon (approved):

PR #179, opened 2024-05-22
Open telemetry failure modes

Modularius:
Pipeline can no longer panic if OpenTelemetry returns an error. Instead an error message is written to stderr by OpenTelemetry. This behaviour can be customised by a subsequent PR.
OpenTelemetry errors occur when
- An invalid URI is given as otel-endpoint.
- The Jaeger collector becomes unavailable and the component attempts to export traces.

If the Jaeger collector becomes available again, then trace exporting resumes seamlessly.

### DIFF
--- a/common/src/tracer.rs
+++ b/common/src/tracer.rs
@@ -1,5 +1,4 @@
 use opentelemetry::{
-    global::Error,
     propagation::{Extractor, Injector},
     trace::TraceError,
 };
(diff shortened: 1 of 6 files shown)


common/src/tracer.rs
```diff
@@ -23,20 +24,102 @@ macro_rules! conditional_init_tracer {
...
+                let tracer = OtelTracer::new(
                     otel_endpoint,
                     env!("CARGO_BIN_NAME"),
                     Some((module_path!(), $level)),
-                )
-                .expect("Open Telemetry Tracer is created")
+                );
             })
             .or_else(|| {
                 tracing_subscriber::fmt::init();
```
DanNixon, line 34: Should there be a warning here that OpenTelemetry was specified but unavailable?

DanNixon: Why not here? You are initialising `tracing_subscriber` which I assume you are only doing because OpenTelemetry is not available, if not here then where?

Modularius: The `or_else` is only called if the user deliberately does not provided a otel-endpoint cli parameter. These lines are not error handling.

DanNixon (approved):

Issue #184, opened 2024-05-30
Should FrameMetadata equality include veto flags?


Modularius:
Digitisers in the same frame have different veto flags.
 - If this is the expected behaviour, then the `FrameMetadata` equality predicate should be changed to ignore `veto_flag`, and the correct method for amalgamating veto flags should be determined (bitwise AND/OR?)
 - Otherwise this should be reported to Dan P and Andrea A as an issue.

PR #186, opened 2024-05-30
Fix veto flag aggregation issue

Modularius:
Digitiser-Aggregator was assigning digitiser messages to different frames because these messages had different veto flags. These modifications fixes that issue, and aggregates the veto flags by bitwise OR.

Closes https://github.com/STFC-ICD-Research-and-Design/supermusr-data-pipeline/issues/184.

### DIFF
--- a/streaming-types/src/frame_metadata.rs
+++ b/streaming-types/src/frame_metadata.rs
@@ -13,6 +13,16 @@ pub struct FrameMetadata {
     pub veto_flags: u16,
 }
 
+impl FrameMetadata {
+    pub fn are_frames_equal(&self, other: &Self) -> bool {
+        self.timestamp == other.timestamp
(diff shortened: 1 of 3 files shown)


streaming-types/src/frame_metadata.rs
```diff
@@ -13,6 +13,16 @@ pub struct FrameMetadata {
     pub veto_flags: u16,
 }
 
+impl FrameMetadata {
+    pub fn are_frames_equal(&self, other: &Self) -> bool {
```
DanNixon, line 17:
```suggestion
    pub fn equals_ignoring_veto_flags(&self, other: &Self) -> bool {
```

DanNixon (changes requested):

DanNixon (approved):

PR #191, opened 2024-06-04
Deduplicate code

### LINKED ISSUE #24: Deduplicate code
Currently there is a lot of duplicated code for things such as live processing of Kafka messages, etc. which should be addressed.

Henry-Denny:
Closes #24.


common/src/lib.rs
```diff
@@ -44,3 +47,28 @@ pub fn generate_kafka_client_config(
     }
     client_config
 }
+
+pub fn create_default_consumer(
+    broker_address: &String,
+    username: &Option<String>,
+    password: &Option<String>,
+    consumer_group: &String,
+    topics_to_subscribe: Option<&[&str]>,
```
Modularius, line 56: Does `topics_to_subscribe: Option<&[&str]>` need to be `Option`? What happens when `consumer.subscribe` is given an empty list? I don't think that case will ever arise anyway as a consumer should always subscribe to at least one topic.

Henry-Denny: That makes sense, I'm not sure why I made it `Option` to begin with.

diagnostics/src/daq_trace/mod.rs
```diff
@@ -79,6 +79,46 @@ impl DigitiserData {
     }
 }
 
+type SharedData = Arc<Mutex<HashMap<u8, DigitiserData>>>;
```
DanNixon, line 82: Are the changes to this file intentional?

Henry-Denny: They are not.

DanNixon (approved):

DanNixon:
@Henry-Denny There is a conflict here.

DanNixon (approved):

Issue #177, opened 2024-05-20
Remove Digitiser Frame option from Nexus Writer


Modularius:
Originally I didn't know whether the nexus writer needed to handle DigitizerEventListMessage as well as FrameAssembledEventListMessage.

To streamline the Nexus Writer, remove event_message.rs and replace instances of GenericEventMessage with FrameAssembledEventListMessage wherever it is used in the rest of the component.

PR #198, opened 2024-06-07
Removed generic event message and digitiser_event_message

Modularius:
Removed digitiser event message functionality from nexus writer, which is not needed.

Closes https://github.com/STFC-ICD-Research-and-Design/supermusr-data-pipeline/issues/177


nexus-writer/src/main.rs
```diff
@@ -65,9 +60,6 @@ struct Cli {
     #[clap(long)]
     alarm_topic: Option<String>,
 
-    #[clap(long)]
-    digitiser_event_topic: Option<String>,
-
     #[clap(long)]
     frame_event_topic: Option<String>,
```
DanNixon, line 64: This should likely no longer be optional.

DanNixon (changes requested):

nexus-writer/src/main.rs
Modularius: Should any of the topics be optional?

DanNixon (approved):

Issue #169, opened 2024-05-03
Refactor aggregator and nexus writer to separate span functionality


Modularius:
The Nexus Writer and Cache structs in nexus-writer and digitiser-aggregator, respectively both contain data which persists over several Kafka messages. For observability, they own a span to represent each run or frame respectively.
We would like to factor out this functionality into a trait (defined in spanned.rs) that each struct must implement, instead of each struct implementing its own ad hoc methods.

This trait would looks something like:
```rust
trait PersistentSpan {
  type Key : PartialEq;

  fn find_span(key: &Key) -> &SpanOnce;
  fn find_span_mut(key: &Key) -> &mut SpanOnce;
}
```
and for instance
```rust
impl PersistentSpan for Run {
  type Key = Metadata;
  ...
}
```

PR #204, opened 2024-06-12
Decouple span

Modularius:
- Added a `FindSpan`/`FindSpanMut` trait (defined in Common/.Spanned.rs) for `FrameCache` in the aggregator to isolate the function extracting the associated span from `PartialFrame`, from the rest of `FrameCache`
- Removed the `root_span` field from `FrameCache` and `NexusEngine`, and set the associated spans of `PartialFrame` and `Run` to `parent: None` instead.
- Converted the above associated spans from `trace_span`s to `info_span`s and assigned them `target: otel`.

Closes https://github.com/STFC-ICD-Research-and-Design/supermusr-data-pipeline/issues/169


digitiser-aggregator/src/frame/cache.rs
```diff
@@ -73,6 +59,18 @@ where
     }
 }
 
+impl<D: Debug> FindSpan for FrameCache<D> {
+    type Key = FrameMetadata;
+}
```
DanNixon, line 64:
```suggestion
}

```

DanNixon (changes requested):

DanNixon (approved):

Issue #202, opened 2024-06-11
Nexus Writer Alterations


Modularius:
- The nexus class for the log items in the runlog and selog groups should be changed to NXlog following the implementation in other Nexus files
- ~~The offsets of selog entries and the event_time_zero fields should be from the start of the run rather than the time of the first incoming selog or data message~~ (This now has its own issue).

Issue #8, opened 2022-08-30
nexus-writer


DanNixon:
Write a Mantid compatible NeXus file given detector, metadata and run boundary information via Kafka.

There are several kinds of message that can be received via Kafka (some may be on different topics):

- Run start/stop
- Detector data
- Metadata

The run start and stop messages are used to control the file writer, i.e. a run start message will contain enough information to start writing the subsequent detector data and metadata to a file, and the run stop message marks the end of this file.

Detector data may be events or histograms, ideally the writer should be able to handle both (although the ability to handle both simultaneously is not required).
Current files are typically histograms.

This task should be broken down into two stages:

- [x] taking an existing file, write functions to build an equivalent NeXus file
- [x] receiving data from Kafka to put into files

See:

- [hdf5](https://docs.rs/hdf5/latest/hdf5/) is the Rust crate to use for writing HDF5 files.
- https://www.nexusformat.org/

Modularius:
Run start/stop and Detector data messages are now able to be processed, metadata messages are currently in progress.

PR #207, opened 2024-06-16
Nexus writer finalisation

Modularius:
- Add capability to customize the chunk sizes of vectors in the nexus file.
- Changed Nexus classes from "NX::RUNLOG" to "NX::LOG" in a couple of instances.

Closes https://github.com/STFC-ICD-Research-and-Design/supermusr-data-pipeline/issues/202 and https://github.com/STFC-ICD-Research-and-Design/supermusr-data-pipeline/issues/8


nexus-writer/src/main.rs
```diff
@@ -82,6 +82,12 @@ struct Cli {
 
     #[clap(long, default_value = "127.0.0.1:9090")]
     observability_address: SocketAddr,
+
+    #[clap(long, default_value = "1048576")]
```
DanNixon, line 86: How was this default selected?

DanNixon (approved):

nexus-writer/src/main.rs
Modularius: This is just 2^20, it seems like a reasonable choice from experience.

PR #214, opened 2024-07-03
Add metrics to all components

### LINKED ISSUE #196: Add metrics to all components
The components digitiser-aggregator and nexus-writer don't use metrics. This is probably because the Kagiyama crate was removed from these crates earlier, so was missed in the change to metrics.rs.

Metrics should be recorded in these crates in a fashion analogous to how it is done in trace-to-events for instance.

Henry-Denny:
Closes #196.


digitiser-aggregator/src/main.rs
```diff
@@ -162,16 +197,27 @@ async fn on_message(
                             .await;
                         }
                         Err(e) => warn!("Invalid Metadata: {e}"),
+                        //  NOTE: Should any metrics be recorded here?
```
DanNixon, line 200: Yes.

digitiser-aggregator/src/main.rs
```diff
@@ -162,16 +197,27 @@ async fn on_message(
...
+                    .increment(1);
                 }
             }
         } else {
             warn!("Unexpected message type on topic \"{}\"", msg.topic());
             debug!("Message: {msg:?}");
             debug!("Payload size: {}", payload.len());
+            counter!(
+                MESSAGES_RECEIVED,
+                &[messages_received::get_label(MessageKind::Unknown)]
```
DanNixon, line 218: This category should probably be called "unexpected" rather than "unknown".

This could be a known message type, but just one that we did not expect to see.

digitiser-aggregator/src/main.rs
```diff
@@ -199,8 +245,20 @@ async fn cache_poll(
                 .send(future_record, Timeout::After(Duration::from_millis(100)))
                 .await
             {
-                Ok(r) => debug!("Delivery: {:?}", r),
-                Err(e) => error!("Delivery failed: {:?}", e),
+                Ok(r) => {
+                    debug!("Delivery: {:?}", r);
+                    //  WARN: Check whether this should count as a message being processed.
+                    counter!(MESSAGES_PROCESSED).increment(1)
```
DanNixon, line 251: No, this does not really make sense. There should be a metric that represents the number of complete frames sent by the aggregator.

In general, avoid trying to pigeon hole metrics into a few common names, create the metrics that make sense in a component first then combine the common ones where appropriate.

DanNixon (changes requested):

common/src/metrics.rs
```diff
@@ -28,21 +39,25 @@ pub mod messages_received {
...
 
     // Label building function
     pub fn get_label(failure_kind: FailureKind) -> (&'static str, &'static str) {
         (
             "failure_kind",
             match failure_kind {
-                FailureKind::UnableToDecodeMessage => "unable_to_decode_message",
                 FailureKind::DataProcessingFailed => "data_processing_failed",
-                FailureKind::KafkaPublishFailed => "kafka_publish_failed",
+                FailureKind::DeliveryFailed => "delivery_failed",
```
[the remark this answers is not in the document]
Henry-Denny: I didn't realise these were the same.

Henry-Denny:
# Current state of PR

Most metrics have been added, however there still might be some additional ones that should be recorded. In particular the NeXus writer still has some gauges which could be added.

DanNixon (approved):

PR #222, opened 2024-07-31
Reduce CI load

DanNixon:
To test:

- Review changes to CI config
- See that formatting always runs (it should have ran on this PR)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,6 +16,19 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
+
+      - uses: dorny/paths-filter@v3
+        id: filter
+        with:
+          filters: |
+            flake:
+              - '**/*.nix'
+              - 'flake.lock'
+            rust:
+              - '**/*.rs'
+              - '**/Cargo.toml'
+              - 'Cargo.lock'
+
       - uses: DeterminateSystems/nix-installer-action@v12
       - uses: DeterminateSystems/magic-nix-cache-action@v7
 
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -28,23 +41,25 @@ jobs:
...
       - name: Clippy
+        if: steps.filter.outputs.rust == 'true'
         run: nix develop --command cargo clippy --all-targets -- -D warnings
 
       - name: Repository cleanliness
+        if: steps.filter.outputs.rust == 'true'
         run: git diff --exit-code
 
       - name: Test
+        if: steps.filter.outputs.flake == 'true' && steps.filter.outputs.rust == 'true'
```
Modularius, line 52: Wouldn't this only run if both rust and nix files have been changed? Maybe it should be
```nix
if: steps.filter.outputs.flake == 'true' || steps.filter.outputs.rust == 'true'
```

DanNixon: Ah, yes. Good catch.

.github/workflows/component.yml
```diff
@@ -0,0 +1,74 @@
...
+              - 'Cargo.lock'
+              # Rust common
+              - 'common/Cargo.toml'
+              - 'common/**/*.rs'
+              - 'streaming-types/Cargo.toml'
+              - 'streaming-types/**/*.rs'
+              # Component
+              - "${{ inputs.component }}/flake.nix"
+              - "${{ inputs.component }}/Cargo.toml"
+              - "${{ inputs.component }}/**/*.rs"
```
Modularius, line 38: Should there also be filters for the default.nix files and README.md in each component?

DanNixon: Ah, that was supposed to be `default.nix` not `flake.nix` there.

The readme is not required as this has no influence over the binary.

.github/workflows/ci.yml
```diff
@@ -28,23 +41,25 @@ jobs:
         run: nix develop --command treefmt --fail-on-change
 
       - name: Clippy
+        if: steps.filter.outputs.rust == 'true'
         run: nix develop --command cargo clippy --all-targets -- -D warnings
 
       - name: Repository cleanliness
+        if: steps.filter.outputs.rust == 'true'
```
Modularius, line 48: Also, I think this step would be needed if nix files had been edited too.

Modularius (approved):

PR #231, opened 2024-08-09
Round 1 of error handling improvements

DanNixon:
## Summary of changes

Initial work to improve error handling.

- general code review
- see that error messages make sense

### DIFF
--- a/diagnostics/src/daq_trace/mod.rs
+++ b/diagnostics/src/daq_trace/mod.rs
@@ -4,7 +4,6 @@ mod ui;
 use self::app::App;
 use self::ui::ui;
 use super::DaqTraceOpts;
-use anyhow::Result;
 use chrono::{DateTime, Utc};
 use crossterm::event::{self, DisableMouseCapture, EnableMouseCapture, Event as CEvent, KeyCode};
 use crossterm::execute;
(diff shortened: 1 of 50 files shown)


diagnostics/src/daq_trace/mod.rs
```diff
@@ -300,7 +312,7 @@ async fn poll_kafka_msg(consumer: StreamConsumer, common_dig_data_map: Digitiser
                     }
                 }
 
-                consumer.commit_message(&msg, CommitMode::Async).unwrap();
+                let _ = consumer.commit_message(&msg, CommitMode::Async);
```
Modularius, line 315: This should probably emit an error trace:
```rust
if let Err(e) = consumer.commit_message(&msg, CommitMode::Async) {
    error!("Failed to commit message consume: {e}");}
```

Modularius (changes requested):

diagnostics/src/daq_trace/mod.rs
DanNixon: Not needed here, since this is a TUI app that will 1) never be connected to an external trace collection tool and, 2) never show log messages on the console.

Modularius (approved):

PR #242, opened 2024-08-23
Removed TIMESTAMP_FORMAT

Modularius:
## Summary of changes

~~Moved the constant TIMESTAMP_FORMAT to common crate from the Nexus-Writer crate. This constant is used to parse `DateTime` objects into strings. Hence it is more logical to have a common format in a logical place.~~

Removed the constant `TIMESTAMP_FORMAT` from nexus-writer, and replaced its usage with `to_rfc3339()`.

Ensure changes to code are as described above.

### DIFF
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -17,6 +17,7 @@ pub type FrameNumber = u32;
 pub type SampleRate = u64;
 
 pub const CHANNELS_PER_DIGITIZER: usize = 8;
+pub const TIMESTAMP_FORMAT: &str = "%Y-%m-%dT%H:%M:%S%.f%z";
 
 pub fn channel_index(digitizer_index: usize, channel_index: usize) -> usize {
     (digitizer_index * CHANNELS_PER_DIGITIZER) + channel_index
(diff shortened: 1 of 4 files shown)


DanNixon:
Is this actually needed outside of `nexus-writer`? If not then it would be better left where it is.

Can we just rename it to something that better describes what it is for (i.e. *which* timestamps is this the format for).

Modularius:
It will be used in other crates in one of the other PRs, (I think in the aggregator).

DanNixon:
In what context will it be used there? Even if the formats are the same we should be explicit about what a specific format is for.

As a side note, the format it describes is just RFC 3339, why do we need the format to be explicitly defined outside of the support already provided by `chrono`?

Modularius:
In nexus-writer is is used to specify the format of timestamps that appear as strings in the file format. At the time, it seemed wise to allow us to be flexible with this, but looking back ISO8601 is the standard.
In other crates, I only use it in OpenTelemetry and tracing, so to_rfc3339() will be more than adequate.

I'll change this PR to remove TIMESTAMP_FORMAT and use to_rfc3339() instead.

DanNixon (approved):

PR #243, opened 2024-08-23
Added SpannedAggregator to common crate

Modularius:
## Summary of changes

Introduced SpannedAggregator trait to encapsulate the span-aggregating behaviour of digitiser_aggregator::Frame and nexus_writer::Run.

Ensure changes are as described.

### DIFF
--- a/common/src/spanned.rs
+++ b/common/src/spanned.rs
@@ -95,26 +95,26 @@ pub trait SpannedMut: Spanned {
 /// Types which have a span: SpanOnce field may implement this trait which
 /// is intended to encapsulate span-aggregating behaviour.
 pub trait SpannedAggregator: SpannedMut {
-    fn span_init(&mut self);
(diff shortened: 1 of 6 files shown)


DanNixon:
On it's own, a trait with no uses or implementations is not really an atomic PR. What will it be used for? Is there a reason there are no implementations?

Modularius:
I wasn't sure how you wanted me to split the big PR up. This will be implemented in nexus writer and aggregator in a future PR. I thought it best to implement them separately to cut the number of "files changed" down. I had thought it might be difficult to piece apart the implementation from the other changes, but in retrospect, it makes more sense to put the basic implementations in here.

Modularius:
Note some changes had to be made to the FindSpan and FindSpanMut traits to get the same functionality, but those traits will disappear in a future PR so the extra complexity will disappear.

PR #244, opened 2024-08-23
Added record_metadata_fields_to_span macro to common crate

Modularius:
## Summary of changes

Added `record_metadata_fields_to_span` macro to populate spans which are setup to record metadata fields.

Does code reflect description above, and does the attached doc comment make sense and adequately explain its purpose.

### DIFF
--- a/digitiser-aggregator/src/main.rs
+++ b/digitiser-aggregator/src/main.rs
@@ -20,6 +20,7 @@ use supermusr_common::{
         messages_received::{self, MessageKind},
         metric_names::{FAILURES, FRAMES_SENT, MESSAGES_PROCESSED, MESSAGES_RECEIVED},
     },
+    record_metadata_fields_to_span,
     spanned::{FindSpanMut, Spanned},
     tracer::{FutureRecordTracerExt, OptionalHeaderTracerExt, TracerEngine, TracerOptions},
     CommonKafkaOpts, DigitizerId,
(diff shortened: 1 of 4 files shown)


DanNixon:
On it's own this isn't really an atomic PR. What calls `record_metadata_fields_to_span`? Is there a reason it is not actually used?

Modularius:
As with the other PR I was going to add the callers in the other crates later, but it makes sense to add them in here wherever possible.

nexus-writer/src/main.rs
```diff
@@ -253,6 +265,13 @@ fn process_frame_assembled_event_list_message(nexus_engine: &mut NexusEngine, pa
     match root_as_frame_assembled_event_list_message(payload) {
         Ok(data) => match nexus_engine.process_event_list(&data) {
             Ok(run) => {
+                let metadata_result: Result<FrameMetadata, _> = data.metadata().try_into();
```
DanNixon, line 268:
```suggestion
                match data.metadata().try_into::<FrameMetadata>();
```

digitiser-aggregator/src/main.rs
```diff
@@ -180,6 +189,9 @@ async fn on_message(
                     let metadata_result: Result<FrameMetadata, _> = msg.metadata().try_into();
```
DanNixon, line 189: As per the other comment.

DanNixon (changes requested):

nexus-writer/src/main.rs
Modularius: `try_into()` doesn't take generics, so it would have to be either
```rust
match TryInto::<FrameMetadata>::try_into(data.metadata()) {
```
or
```rust
match data.metadata().try_into() as Result<FrameMetadata,_> {
````
I think the second would be better, but not sure which one you'd prefer as standard.

DanNixon (approved):

PR #248, opened 2024-09-04
Updates to digitiser aggregator, and adding fields to spans

Modularius:
### cache.rs
1. ~~Swapped `VecDeque` to `HashMap<FrameMetadata,_>` and implemented resulting changes~~
2. Added metadata to `push` function instrumentation.
3. Added `get_num_partial_frames` function.
4. Added `test_metadata_equality` test.
5. ~~`push` now returns `&impl SpannedAggregator`, and dependence on `FindSpan(Mut)` removed.~~

### partial.rs
1. Added metadata to `Frame` span.

### main.rs
1. Refactored `on_message` to `process_kafka_message` and `process_digitiser_event_list_message` functions.
2. Added metadata and other fields to `process_kafka_message` and `process_digitiser_event_list_message` instrumentation
3. Instrumented `root_as_digitizer_event_list_message` by wrapping with `span_of_root_as_digitizer_event_list_message`

### frame_metadata.rs
1. ~~Implemented `Hash` trait for `FrameMetadata`~~
2. Implemented `PartialEq` for `FrameMetadata` as a temporary measure to exclude `veto_flags` (this will be removed later).

1. Do fields added to spans conform with the rules set out in `tracing.md`?
2. General code review.

### DIFF
--- a/digitiser-aggregator/src/frame/cache.rs
+++ b/digitiser-aggregator/src/frame/cache.rs
@@ -45,8 +45,9 @@ where
             .or_insert_with(|| {
                 // or create a new PartialFrame
                 let mut frame = PartialFrame::<D>::new(self.ttl, metadata.clone());
+
+                // Initialise the span field
                 if let Err(e) = frame.span_init() {
-                    // Initialise the span field
                     warn!("Frame span initiation failed {e}")
                 }
                 frame
@@ -235,6 +236,8 @@ mod test {
             veto_flags: 5,
         };
 
+        assert_eq!(frame_1, frame_2);
+
         assert_eq!(cache.frames.len(), 0);
         assert!(cache.poll().is_none());
 
@@ -245,6 +248,5 @@ mod test {
         cache.push(2, &frame_2, EventData::dummy_data(0, 5, &[0, 1, 2]));
         assert_eq!(cache.frames.len(), 1);
         assert!(cache.poll().is_some());
-        //let frame = cache.poll().is_some();
     }
 }
--- a/digitiser-aggregator/src/frame/cache.rs
+++ b/digitiser-aggregator/src/frame/cache.rs
@@ -60,7 +60,7 @@ where
 
                     frame.push(digitiser_id, data);
                     self.frames.push_back(frame);
-                    self.frames.back().expect("Partial Frame Exists")
+                    self.frames.back().expect("This should never fails")
                 }
             }
         };
@@ -90,7 +90,7 @@ where
             .front()
             .is_some_and(|frame| frame.is_complete(&self.expected_digitisers) | frame.is_expired())
         {
-            let frame = self.frames.pop_front().expect("Frame Exists");
+            let frame = self.frames.pop_front().expect("This should never fail");
             if let Err(e) = frame.end_span() {
                 warn!("Frame span drop failed {e}")
             }
--- a/digitiser-aggregator/src/frame/cache.rs
+++ b/digitiser-aggregator/src/frame/cache.rs
@@ -1,6 +1,6 @@
 use super::{partial::PartialFrame, AggregatedFrame};
 use crate::data::{Accumulate, DigitiserData};
-use std::{collections::HashMap, fmt::Debug, time::Duration};
+use std::{collections::VecDeque, fmt::Debug, time::Duration};
 use supermusr_common::{spanned::SpannedAggregator, DigitizerId};
 use supermusr_streaming_types::FrameMetadata;
 use tracing::warn;
@@ -9,7 +9,7 @@ pub(crate) struct FrameCache<D: Debug> {
     ttl: Duration,
     expected_digitisers: Vec<DigitizerId>,
 
-    frames: HashMap<FrameMetadata, PartialFrame<D>>,
+    frames: VecDeque<PartialFrame<D>>,
 }
 
 impl<D: Debug> FrameCache<D>
@@ -39,10 +39,12 @@ where
         metadata: &FrameMetadata,
         data: D,
     ) -> &'a impl SpannedAggregator {
-        let frame = self
+        if self
             .frames
-            .entry(metadata.clone()) // Find the frame with the given metadata
-            .or_insert_with(|| {
+            .iter()
+            .all(|frame| frame.metadata != *metadata)
+        {
+            self.frames.push_back({
                 // or create a new PartialFrame
                 let mut frame = PartialFrame::<D>::new(self.ttl, metadata.clone());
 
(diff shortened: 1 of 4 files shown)


digitiser-aggregator/src/frame/cache.rs
```diff
@@ -197,4 +211,40 @@ mod test {
...
+        assert!(cache.poll().is_none());
+
+        cache.push(1, &frame_1, EventData::dummy_data(0, 5, &[0, 1, 2]));
+        assert_eq!(cache.frames.len(), 1);
+        assert!(cache.poll().is_none());
+
+        cache.push(2, &frame_2, EventData::dummy_data(0, 5, &[0, 1, 2]));
+        assert_eq!(cache.frames.len(), 1);
+        assert!(cache.poll().is_some());
+        //let frame = cache.poll().is_some();
```
DanNixon, line 248: Is this commented out for a reason?

digitiser-aggregator/src/frame/cache.rs
```diff
@@ -27,52 +24,66 @@ where
...
+            .entry(metadata.clone()) // Find the frame with the given metadata
+            .or_insert_with(|| {
+                // or create a new PartialFrame
                 let mut frame = PartialFrame::<D>::new(self.ttl, metadata.clone());
-                frame.push(digitiser_id, data);
-                self.frames.push_back(frame);
-            }
-        }
+                if let Err(e) = frame.span_init() {
+                    // Initialise the span field
```
DanNixon, line 49: This comment feels like it is in the wrong position.

streaming-types/src/frame_metadata.rs
```diff
@@ -23,6 +24,26 @@ impl FrameMetadata {
...
+}
+
+impl Hash for FrameMetadata {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        self.timestamp.hash(state);
+        self.period_number.hash(state);
+        self.protons_per_pulse.hash(state);
+        self.running.hash(state);
+        self.frame_number.hash(state);
+        // At the moment we do no consider the veto_flags
```
DanNixon, line 42:
```suggestion
        // At the moment we do not consider the veto_flags
```

DanNixon (changes requested):

digitiser-aggregator/src/frame/cache.rs
Modularius: Indeed, it is meant to apply to the above line. I think treefmt moved it from the end of line 48 to its own line.

DanNixon:
> Swapped VecDeque to HashMap<FrameMetadata,_> and implemented resulting changes

Why was this done?

Modularius:
It was the cleanest way to make the `push` function return a pointer to the `PartialFrame` object, plus it seems a more natural container to use for matching on `FrameMetadata`.

DanNixon:
The issue is that in doing so you remove the deterministic ordering of frames emitted from the aggregator in the case where multiple frames are held in the cache.

Modularius:
I see, perhaps we can keep a `VecDeque` with the `FrameMetadata` so they can be emitted in sequence. I'll have another look at reimplementing the `VecDeque` solution first though.

DanNixon:
I think I am missing what was wrong with using `VedDeque` as it was.

Modularius:
I was tripping over the borrow checker trying to return a ref to its contents, whilst also possibly modifying it if a new `PartialFrame` was needed

DanNixon:
> I was tripping over the borrow checker trying to return a ref to its contents, whilst also possibly modifying it if a new `PartialFrame` was needed

I assume the only reason you need to return anything from `push()` is to handle span linking [here](https://github.com/STFC-ICD-Research-and-Design/supermusr-data-pipeline/pull/248/files#diff-63e4f8e251e38c8993252a0794e497e696c0698fba7685a491746742e45b58a1R253)?
I think this makes more sense to be handled in `push()` itself.

Modularius:
That's right. I prefer to return from `push` to decouple as much of the Otel handling from the rest of the functionality by keeping it in `main.rs`.

I managed to find a way to reimplement `VecDeque` though.

digitiser-aggregator/src/frame/cache.rs
```diff
@@ -27,52 +24,64 @@ where
...
+    #[tracing::instrument(skip_all, fields(
+        digitiser_id = digitiser_id,
+        metadata_timestamp = metadata.timestamp.to_rfc3339(),
+        metadata_frame_number = metadata.frame_number,
+        metadata_period_number = metadata.period_number,
+        metadata_veto_flags = metadata.veto_flags,
+        metadata_protons_per_pulse = metadata.protons_per_pulse,
+        metadata_running = metadata.running
+    ))]
+    pub(crate) fn push<'a>(
```
DanNixon, line 36: Not a fan of this approach, as it introduces an exhaustive search of the cache that is not required.

I don't think the strong separation of the tracing specific code is helpful here, in fact I think the readability of the code suffers if return values exist only to facilitate observability.

By its very nature, observability code does end up in more than one part of a codebase, so keeping it all in `main.rs` is an anti-pattern best avoided IMO.

DanNixon (changes requested):

digitiser-aggregator/src/frame/cache.rs
Modularius: Ah true enough. I assumed we'd be unlikely to have a situation where more than two frames would be in the cache at a time. I suppose that could happen if a digitiser event message goes missing

digitiser-aggregator/src/frame/cache.rs
```diff
@@ -27,52 +24,84 @@ where
...
+                    let mut frame = PartialFrame::<D>::new(self.ttl, metadata.clone());
+
+                    // Initialise the span field
+                    if let Err(e) = frame.span_init() {
+                        warn!("Frame span initiation failed {e}")
+                    }
+
+                    frame.push(digitiser_id, data);
+                    self.frames.push_back(frame);
+                    self.frames.back().expect("Partial Frame Exists")
```
DanNixon, line 63: `expect()` strings should contain the word "should", the panic message this generates would be ambiguous about if the error is the existence or absence of a partial frame.

I have also been adding ", this should never fail" to panic messages that are by (and only by) guarantees of the APIs used and the way they are used never going to fail.

DanNixon (changes requested):

digitiser-aggregator/src/frame/cache.rs
DanNixon: I did not mean entirely replace the message with "this should never fail", the format should be "\<assertion including the word should\>, this should never fail".

Modularius: Sorry, got ya now.

DanNixon (approved):
:+1:

PR #252, opened 2024-09-13
Updated Simulator Spans and Error Handling

Modularius:
## Summary of changes

1. Updated Simulator spans to conform with OpenTelemetry policy.
2. Updated error handling to conform with new requirements.
3. Added `digitiser_` prefix to `trace` and `event` cli options to better describe their purpose.

Ensure spans and error handling conforms to policies.

### DIFF
--- a/simulator/src/integrated/mod.rs
+++ b/simulator/src/integrated/mod.rs
@@ -5,7 +5,7 @@ pub(crate) mod simulation;
 pub(crate) mod simulation_elements;
 pub(crate) mod simulation_engine;
 
-use crate::{Cli, Defined};
+use crate::Defined;
 use rdkafka::producer::FutureProducer;
 use simulation::{Simulation, SimulationError};
 use simulation_engine::{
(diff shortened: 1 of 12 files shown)


simulator/src/integrated/mod.rs
```diff
@@ -50,17 +57,30 @@ pub(crate) async fn run_configured_simulation(
             producer,
             kafka_producer_thread_set: &mut kafka_producer_thread_set,
             topics: Topics {
-                traces: cli.trace_topic.as_deref().unwrap(),
-                events: cli.event_topic.as_deref().unwrap(),
-                frame_events: cli.frame_event_topic.as_deref().unwrap(),
-                run_controls: cli.control_topic.as_deref().unwrap(),
+                traces: cli
```
DanNixon, line 60: This error checking would be better handled via `clap`.

DanNixon (changes requested):

simulator/src/integrated/mod.rs
Modularius: Running the simulator in different modes (i.e. cli subcommands) makes use of different combinations of these topics. For instance, they aren't all required for sending Run Commands, or emitting a simulated Trace using the original `Single` subcommand.

DanNixon: You can still handle this with `clap`, [this](https://github.com/DanNixon/emfcamp-schedule-api/blob/main/cli/src/commands/upcoming.rs#L10) is an example of solving pretty much the same problem. i.e. you have several commands that use different combinations of the same arguments.

DanNixon (approved):

PR #254, opened 2024-09-24
Add testing utility crate

DanNixon:
## Summary of changes

Adds a barebones testing utility crate (mostly stolen from Satori).

- review code, plz

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -207,7 +207,7 @@ dependencies = [
  "pin-project-lite",
  "rustversion",
  "serde",
- "sync_wrapper",
+ "sync_wrapper 0.1.2",
  "tower",
  "tower-layer",
  "tower-service",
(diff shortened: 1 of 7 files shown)


Modularius:
Did some minor error handling improvements.

Is this intended to be called from a yet to be made binary? Or will this crate eventually be converted into one? In the former case the functions should return Results rather than panic before I approve, but in the latter case, I can approve as I imagine error handling would be done later.

DanNixon:
> Did some minor error handling improvements.

Since this code will only ever be used within tests it is technically exempt from the error handling style rules, although the changes are still for the better.
 
> Is this intended to be called from a yet to be made binary?

It will be used by a suite of integration tests. It does not necessarily need to be it's own crate, that is just the route I went with in Satori.
On second consideration it is probably a better idea to just have this code inside the crate that will contain the tests.

PR #259, opened 2024-10-02
Diagnostic refactoring

Modularius:
## Summary of changes

1. General refactoring to make code more idiomatic
2. Moved the constructor and display functionality of `DigitiserData` to its own file.
3. Reworked the column structure to group up related items (as described in the README).
4. Added `weights` options to the column widths, making some columns larger than others.
5. Introduced `Channel` column which monitors a single channel (adjustable with left/right keys) and displays channel id and sample range.

General code review.
Test with command such as
```bash
cargo run --bin diagnostics daq-trace --broker 130.246.55.29:9092 --topic daq-traces-in --group vis-3
```
when a run is in progress from the chosen broker.


DanNixon (approved):
:+1:

Issue #262, opened 2024-10-07
Enable Nexus-Writer to Move the Nexus File once it has been Completed


Modularius:
## Problem this feature will solve

Nexus Writer takes a single location, creates and writes to a nexus file in real-time, at that location. Once writing is completed, the file is left in that location. This means the long-term location of the file must be the same as short-term location whilst data is being written.
This is a problem because long-term storage is usually not optimised for fast disk read/writes.

## Description of solution

Allow the user to specify an `archive path` in the cli, a location to which the file is moved once the writer determines the run is complete. This path is in addition to the current `path` attribute, where the file is located whilst the run is in progress. The file at `path` is deleted as soon as it is transferred to the `archive path`.

Implementation of this feature will also decide how to handle file copy errors.

PR #263, opened 2024-10-14
Add nexus writer move completed file feature

Modularius:
## Summary of changes

Added a feature to nexus-writer to allow the nexus file to be moved from a temporary location to another, when the run is determined to be completed by nexus-writer.
The move is performed async to prevent other runs from being blocked.

General code quality review (full testing requires pipeline simulation).

Closes https://github.com/STFC-ICD-Research-and-Design/supermusr-data-pipeline/issues/262.

### DIFF
--- a/nexus-writer/src/nexus/engine.rs
+++ b/nexus-writer/src/nexus/engine.rs
@@ -20,6 +20,7 @@ pub(crate) struct NexusEngine {
     run_cache: VecDeque<Run>,
     run_number: u32,
     nexus_settings: NexusSettings,
+    run_move_cache: Vec<Run>,
 }
 
 impl NexusEngine {
@@ -30,6 +31,7 @@ impl NexusEngine {
             run_cache: Default::default(),
             run_number: 0,
             nexus_settings,
+            run_move_cache: Default::default(),
         }
     }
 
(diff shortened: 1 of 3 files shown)


nexus-writer/src/nexus/engine.rs
```diff
@@ -170,6 +170,14 @@ impl NexusEngine {
                 if let Err(e) = run.end_span() {
                     warn!("Run span drop failed {e}")
                 }
+                if let Some((file_name, archive_name)) = Option::zip(
+                    self.filename.as_ref(),
+                    self.nexus_settings.archive_path.as_ref(),
+                ) {
+                    if let Err(e) = run.move_to_archive(file_name, archive_name) {
```
DanNixon, line 177: You are not `await`ing the `JoinHandle`, so there is no guarantee the move ever happens.

DanNixon (changes requested):

nexus-writer/src/nexus/engine.rs
Modularius: Wouldn't that only be a problem if the program exits unexpectedly? As it is running continuously the move should always happen in the background, shouldn't it?

DanNixon: Realistically, yes, theoretically, no.
Plus it failing just because the writer exits is not a good design (keep in mind that the NeXus writer might be recycled between runs if the pipeline configuration changes).

PR #273, opened 2024-11-21
Resume Run Functionality

Modularius:
## Summary of changes

Added behaviour for the Nexus Writer to resume a run if it was interrupted by the component going down.
- ~~A file with extension ".partial_run" is added to the local path when a run_start is received. This file is removed when the nexus file is transferred to the archive (signalling a successful run).~~
- When the nexus writer starts, any ~~".partial_run"~~ ".nxs" files in the local path are read and runs re-created (as ~~.partial_run~~ .nxs files indicate that a previous instance of nexus writer had failed).

General code review.

Tool was tested with simulated data.


Modularius:
I considered that, but thought it would couple the design of the file format with this functionality. I'd rather avoid that if possible.

DanNixon:
> I considered that, but thought it would couple the design of the file format with this functionality. I'd rather avoid that if possible.

But all the information is already recorded in the data file (and can be assumed to always be there, as it is all pretty basic information). I can't see why just using that is not the most straightforward.

Modularius:
On second thoughts, you are right that this is the cleanest way to do it.

nexus-writer/src/nexus/run_parameters.rs
```diff
@@ -103,4 +105,11 @@ impl RunParameters {
             params.last_modified = Utc::now();
         }
     }
+
+    pub(crate) fn get_hdf5_path_buf(path: &Path, run_name: &str) -> PathBuf {
```
DanNixon, line 109:
```suggestion
    pub(crate) fn get_hdf5_filename(path: &Path, run_name: &str) -> PathBuf {
```

nexus-writer/src/nexus/run.rs
```diff
@@ -126,16 +126,15 @@ impl Run {
...
-        filename: Option<&Path>,
+        local_path: Option<&Path>,
         message: &FrameAssembledEventListMessage,
     ) -> anyhow::Result<()> {
-        if let Some(filename) = filename {
-            let mut hdf5 = RunFile::open_runfile(filename, &self.parameters.run_name)?;
-            hdf5.push_message_to_runfile(&self.parameters, message)?;
+        if let Some(local_path) = local_path {
+            let mut hdf5 = RunFile::open_runfile(local_path, &self.parameters.run_name)?;
+            hdf5.push_message_to_runfile(message)?; //&self.parameters,
```
DanNixon, line 134: Is the comment relevant?

nexus-writer/Test1.nxs
DanNixon, line 1: What is this file for?

nexus-writer/TestRun1.nxs
DanNixon, line 1: What is this file for?

DanNixon (changes requested):

nexus-writer/Test1.nxs
Modularius: These seem to have been created by the test module at some point. That was not supposed to happen of course, and doesn't happen any more. I think it was an artefact of an alternative use of file handling I was trying (but then abandoned).

PR #277, opened 2024-11-25
Missing run stop prompts new run

Modularius:
## Summary of changes

Adds "emergency_stop" function, so that when a RunStart is encountered whilst a run is currently running, the current run is halted and a new run begins.
If the new RunStart message has the same run_name field as the previous, this results in the previous run being overwritten.

General code review.

Behaviour has been tested with simulated runs.

### DIFF
--- a/nexus-writer/src/nexus/engine.rs
+++ b/nexus-writer/src/nexus/engine.rs
@@ -13,7 +13,7 @@ use supermusr_streaming_types::{
     ecs_f144_logdata_generated::f144_LogData, ecs_pl72_run_start_generated::RunStart,
     ecs_se00_data_generated::se00_SampleEnvironmentData,
 };
-use tracing::{warn, warn_span};
+use tracing::warn;
 
 pub(crate) struct NexusEngine {
     filename: Option<PathBuf>,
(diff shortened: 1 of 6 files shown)


nexus-writer/src/nexus/engine.rs
```diff
@@ -103,27 +103,34 @@ impl NexusEngine {
...
-            )?;
-            if let Err(e) = run.span_init() {
-                warn!("Run span initiation failed {e}")
-            }
-            self.run_cache.push_back(run);
-            Ok(self.run_cache.back_mut().expect("Run exists"))
-        } else {
-            Err(anyhow::anyhow!("Unexpected RunStart Command."))
+        //  If a run is already in progress, and is missing a run-stop
+        //  then call an emergency stop on the current run.
```
DanNixon, line 107: We should not introduce new terminology here. This is already known as an "aborted run".

nexus-writer/src/nexus/run_parameters.rs
```diff
@@ -65,6 +65,20 @@ impl RunParameters {
         }
     }
 
+    #[tracing::instrument(skip_all, level = "trace", err(level = "warn"))]
+    pub(crate) fn set_emergency_stop(&mut self) -> anyhow::Result<()> {
+        if self.run_stop_parameters.is_some() {
+            anyhow::bail!("RunStop already set");
```
DanNixon, line 71: https://github.com/STFC-ICD-Research-and-Design/supermusr-data-pipeline/blob/main/docs/style.md#crab-error-handling-rules

DanNixon (changes requested):

nexus-writer/src/nexus/run_parameters.rs
Modularius: In the "nexus writer refactor" issue I'm replacing all the non-main uses of `anyhow`, so am using that crate with the error handling to be consistent with the current usage.

DanNixon (approved):

PR #279, opened 2024-11-25
Otel Tag Option

Modularius:
## Summary of changes

Adds extra command line option, which prepends the "service.name" field for OpenTelemetry with an optional string. Useful for distinguishing telemetry from different instances of the pipeline running in parallel.

General code review.

Tested on simulated data.

### DIFF
--- a/nexus-writer/src/main.rs
+++ b/nexus-writer/src/main.rs
@@ -98,9 +98,9 @@ struct Cli {
     #[clap(long, default_value = "info")]
     otel_level: LevelFilter,
 
-    /// All OpenTelemetry Tags are appended with this string, if given. Can be used to track different instances of the pipeline running in parallel.
-    #[clap(long)]
-    otel_namespace: Option<String>,
+    /// All OpenTelemetry spans are emitted with this as the "service.namespace" property. Can be used to track different instances of the pipeline running in parallel.
+    #[clap(long, default_value = "")]
+    otel_namespace: String,
 
     /// Endpoint on which OpenMetrics flavour metrics are available
     #[clap(long, default_value = "127.0.0.1:9090")]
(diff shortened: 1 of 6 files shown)


DanNixon:
Is there a reason this modifies the service name rather than adding a new field/tag?

e.g. in the Prometheus world you would typically use a label for the name of the thing that is running and one or more to appropriately designate the specific instance of it.

Modularius:
I initially thought of doing it that way, but I couldn't find a way of doing it that wouldn't be adding a tag to every single span. It might be that there is a way of doing this but this seemed like the most convenient thing to do in the meantime.

Modularius:
Nothing. I think they are used for 3rd party tools like Jaeger rather than as user data.

DanNixon:
`service.instance.id` seems to be what you want: https://opentelemetry.io/docs/specs/semconv/attributes-registry/service/

Modularius:
Ah thanks, yes that would do. I think `service.namespace` is even better for this purpose.

nexus-writer/src/main.rs
```diff
@@ -115,13 +119,14 @@ struct Cli {
 async fn main() -> anyhow::Result<()> {
     let args = Cli::parse();
 
+    debug!("{args:?}");
```
DanNixon, line 122: Should this be here?

nexus-writer/src/main.rs
```diff
@@ -98,6 +98,10 @@ struct Cli {
     #[clap(long, default_value = "info")]
     otel_level: LevelFilter,
 
+    /// All OpenTelemetry Tags are appended with this string, if given. Can be used to track different instances of the pipeline running in parallel.
```
DanNixon, line 101: Do the doc comments need to be changed too?

DanNixon (changes requested):

nexus-writer/src/main.rs
Modularius: That's a relic from past logging, I'll remove it in a nexus-writer specific PR.

DanNixon (approved):

PR #280, opened 2024-11-25
Fixes flushed runs index bug

Modularius:
## Summary of changes

Bug fix.

In very particular circumstances a vector is accessed out of bounds.
When two runs are present in the run cache, and complete at the same time the `flush` method will likely try to access the `run_cache` deque out of bounds.

This bug was missed due to non-idiomatic code being used. It has been replaced and a suitable test added to check this (albeit unlikely) use case.

General code review.

Tested on simulated data.

### DIFF
--- a/nexus-writer/src/nexus/engine.rs
+++ b/nexus-writer/src/nexus/engine.rs
@@ -418,21 +418,21 @@ mod test {
         let ts_start: DateTime<Utc> = GpsTime::new(0, 1, 0, 0, 15, 0, 0, 0).try_into().unwrap();
         let ts_end: DateTime<Utc> = GpsTime::new(0, 1, 0, 0, 17, 0, 0, 0).try_into().unwrap();
 
-        let start = create_start(&mut fbb, "Test1", ts_start.timestamp_millis() as u64).unwrap();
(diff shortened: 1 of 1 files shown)


DanNixon (approved):

nexus-writer/src/nexus/engine.rs
```diff
@@ -174,24 +174,20 @@ impl NexusEngine {
...
-                .run_cache
-                .remove(index)
-                .expect("Index should be within bounds");
-            if let Err(e) = run.end_span() {
-                warn!("Run span drop failed {e}")
+        // Moves the runs into a new vector, then consumes it,
+        // directing completed runs to self.run_move_cache
+        // and incomplete ones back to self.run_cache
+        let temp: Vec<_> = self.run_cache.drain(..).collect();
+        //let mut temp = VecDeque::<Run>::new();
```
DanNixon, line 181: Should this line still be here?

DanNixon (changes requested):

nexus-writer/src/nexus/engine.rs
Modularius: Thanks. No, it shouldn't.

DanNixon (approved):

PR #281, opened 2024-11-27
Changes structure of `Frame Completed` span in nexus writer

Modularius:
## Summary of changes

- In `cache_poll`, the `Frame Completed` span now occurs for every frame dispatched from the cache. Allowing us to see how many are unqueued at the same time.
- Fixes a bug whereby complete frames in the queue are marked as expired because they were behind incomplete frames.
- Removed `debug!("{args}");` log statement.

General code review.

Tested on simulated data.

### DIFF
--- a/digitiser-aggregator/src/frame/cache.rs
+++ b/digitiser-aggregator/src/frame/cache.rs
@@ -83,7 +83,7 @@ where
         if self
             .frames
             .front()
-            .is_some_and(|frame| frame.is_complete() | frame.is_expired())
+            .is_some_and(|frame| frame.is_complete() || frame.is_expired())
         {
             let frame = self
                 .frames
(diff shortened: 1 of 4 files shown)


digitiser-aggregator/src/frame/cache.rs
```diff
@@ -82,7 +83,7 @@ where
         if self
             .frames
             .front()
-            .is_some_and(|frame| frame.is_complete(&self.expected_digitisers) | frame.is_expired())
+            .is_some_and(|frame| frame.is_complete() | frame.is_expired())
```
DanNixon, line 86: Prefer the logical operators when doing logical operations.

```suggestion
            .is_some_and(|frame| frame.is_complete() || frame.is_expired())
```

digitiser-aggregator/src/frame/partial.rs
```diff
@@ -92,7 +101,7 @@ impl<D> SpannedAggregator for PartialFrame<D> {
         #[cfg(not(test))] //   In test mode, the frame.span() are not initialised
         self.span()
             .get()?
-            .record("frame_is_expired", self.is_expired());
+            .record("frame_is_expired", self.is_expired() & !self.is_complete());
```
DanNixon, line 104:
```suggestion
            .record("frame_is_expired", self.is_expired() && !self.is_complete());
```

DanNixon (approved):

PR #282, opened 2024-11-28
Need to Join Completed Producer Threads in Trace to Events

Modularius:
## Summary of changes

Producer threads created in the `trace-to-events` create were not being joined with the main thread, possibly causing excessive memory usage.

This PR introduces a `tokio::select` statement which joins completed threads without waiting for them.

General code review.

Tested on simulated and hifi input.

### DIFF
--- a/trace-to-events/src/main.rs
+++ b/trace-to-events/src/main.rs
@@ -138,18 +138,30 @@ async fn main() -> anyhow::Result<()> {
     let mut kafka_producer_thread_set = JoinSet::new();
 
     loop {
-        match consumer.recv().await {
-            Ok(m) => {
-                process_kafka_message(
-                    &tracer,
-                    &args,
-                    &mut kafka_producer_thread_set,
-                    &producer,
-                    &m,
-                );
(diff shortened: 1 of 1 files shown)


trace-to-events/src/main.rs
```diff
@@ -138,18 +138,30 @@ async fn main() -> anyhow::Result<()> {
...
+                        &mut kafka_producer_thread_set,
+                        &producer,
+                        &m,
+                    );
+                    consumer.commit_message(&m, CommitMode::Async).unwrap();
+                }
+                Err(e) => warn!("Kafka error: {}", e)
+            },
+            r = kafka_producer_thread_set.join_next() => {
+                if let Some(r) = r {
```
DanNixon, line 156: You can simplify this to something like `if let Some(Err(e)) = r`.
Also a slightly more descriptive error message would be useful, in this case provide some context.

DanNixon (changes requested):

DanNixon (approved):

PR #285, opened 2024-12-05
Single producer thread event formation

Modularius:
## Summary of changes

The current event-formation component creates a new thread to produce each outgoing message resulting in the overhead associated with thread for each message. In this PR, the component creates a single persistent thread for producing messages to Kafka, and transfers processed traces from the main thread to this thread via a channel.

Commentary on the new architecture is welcome. Is this a good change?

General code review.

Has been tested on simulated and Kafka data.

### DIFF
--- a/trace-to-events/src/main.rs
+++ b/trace-to-events/src/main.rs
@@ -33,7 +33,7 @@ use supermusr_streaming_types::{
     FrameMetadata,
 };
 use tokio::sync::mpsc::{error::TrySendError, Receiver, Sender};
-use tracing::{debug, error, instrument, metadata::LevelFilter, trace, warn};
+use tracing::{debug, error, info, instrument, metadata::LevelFilter, trace, warn};
 
 #[derive(Debug, Parser)]
 #[clap(author, version, about)]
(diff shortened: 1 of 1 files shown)


trace-to-events/src/main.rs
```diff
@@ -135,7 +140,8 @@ async fn main() -> anyhow::Result<()> {
         "Number of failures encountered"
     );
 
-    let mut kafka_producer_thread_set = JoinSet::new();
+    //let mut kafka_producer_thread_set = JoinSet::new();
```
DanNixon, line 143:
```suggestion
```

trace-to-events/src/main.rs
```diff
@@ -144,18 +150,53 @@ async fn main() -> anyhow::Result<()> {
...
+                    error!("{:?}", e);
+                    counter!(
+                        FAILURES,
+                        &[failures::get_label(FailureKind::KafkaPublishFailed)]
+                    )
+                    .increment(1);
                 }
+            },
+            None => {
+                error!("Send-Frame Receiver Error");
```
DanNixon, line 198:
```suggestion
                info!("Send-Frame channel closed");
```

The channel closing is not an error, it will happen as part of the normal shutdown procedure.

DanNixon (approved):

PR #286, opened 2024-12-11
Close hdf5 file when resuming

Modularius:
## Summary of changes

Fixes two possible source of memory leakage:
 - Explicitly closes the hdf5 file after resuming a partial run
 - Explicitly closes the hdf5 file after an error occurs whilst reading or creating the datasets.

- General code review.
- Advice on limiting unnecessary memory usage.


DanNixon (approved):
:+1:

PR #291, opened 2025-01-08
Digitizer reporting in assembled messages

DanNixon:
## Summary of changes

Adds two extra fields to the assembled frame flatbuffer messages, one to indicate if the frame is considered complete and another to list the digitisers that contributed data to this frame.

This data is already collected in the building of the partial frame, this just adds propagation of it to the assembled message.

- Test using simulated data
- Pipeline should work as it always did


Modularius (approved):
Tested on simulated data, with both complete, and incomplete frames.

Issue #288, opened 2024-12-12
Possible Resolutions to Out-of-Order frames in the nexus files 


Modularius:
## Description of task

If frames appearing out of temporal order in the nexus file proves problematic, a resolution should be decided upon. Possible solutions include:

|Option|Pros|Cons|
|-|-|-|
|Discarding any incomplete frames at the aggregator stage| Easy to implement | If a single digitiser fails to transmit for an extended period, all frames would be lost during this fault |
|Maintain a `latest timestamp` variable and discard any digitiser messages that appears earlier, and does not belong to a currently cached partial frame| Pretty easy to implement | Some data loss |
|Add a pipeline component after the nexus writer that checks for out-of-order frames and reorders data as necessary| No data loss | Potentially difficult to implement, may not be feasible in real-time |
|Integrate aggregator and nexus writer| No data loss | Most difficult to implement, same problems as above, and loss of robustness|

DanNixon:
Option 2 is the best IMO.

- The rate of what would be lost frames is already pretty low, and there is work in progress to further reduce this.
- 3 and 4 introduce too much complexity and erode the modularity of the current design.
- The con for option 1 is a non-starter.

Modularius:
# A Note on Option Two

Option Two prevents frames being split between two frame messages, hence preventing the second frame message being out of order (this would occur if one or more digitisers delayed). However, this does not prevent out of order messages occurring if for two distinct frames the initial digitiser messages arrived out of order. Hence as a solution, option two requires that we assume initial digitiser messages arrive in chronological order.

This is an improvement on the current assumption (that all digitiser messages are chronological and propagate through the pipeline without significant delay) but is worth noting all the same.

PR #298, opened 2025-01-14
Prevent out of order frames

Modularius:
## Summary of changes

In the aggregator: prevents new frames from being created whose timestamp is earlier than or equal to any that have previously been dispatched. That is, if a digitiser messages arrives late, then if it's frame is still in the cache, it is aggregated into that frame, otherwise it is discarded and a warning emitted.

Closes https://github.com/STFC-ICD-Research-and-Design/supermusr-data-pipeline/issues/288.

Note that we still assume that each frame's initial digitiser message appears in chronological order.

General code review.
Has been tested on simulated data.

### DIFF
--- a/digitiser-aggregator/Cargo.toml
+++ b/digitiser-aggregator/Cargo.toml
@@ -6,6 +6,7 @@ edition.workspace = true
 
 [dependencies]
 anyhow.workspace = true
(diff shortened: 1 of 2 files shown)


DanNixon (approved):
:+1:

PR #299, opened 2025-01-17
Upgrade Rust version to 1.84

DanNixon:
## Summary of changes

Moves to Rust 1.84.
Something higher than 1.80 was required to move to the maintained fork of `hdf5`.

This also addresses a few new warnings generated.

- basic test with simulated data
- review modified code

### DIFF
--- a/trace-to-events/src/pulse_detection/datatype/tracepoint.rs
+++ b/trace-to-events/src/pulse_detection/datatype/tracepoint.rs
@@ -6,21 +6,21 @@ use super::{eventdata::Empty, EventData, Temporal, TraceValue};
 pub(crate) trait TracePoint: Clone {
     /// The type which represents the time of the data point.
     /// This should be trivially copyable (usually a scalar).
-    type TimeType: Temporal;
+    type Time: Temporal;
 
     /// The type which contains the value of the data point.
-    type ValueType: TraceValue;
+    type Value: TraceValue;
 
-    type DataType: EventData;
+    type Data: EventData;
 
     /// Returns the time of the data point.
-    fn get_time(&self) -> Self::TimeType;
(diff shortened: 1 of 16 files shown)


trace-to-events/src/pulse_detection/datatype/tracepoint.rs
```diff
@@ -3,25 +3,23 @@ use super::{eventdata::Empty, EventData, Temporal, TraceValue};
 /// An abstraction of the types that are processed by the various filters
 /// To implement TracePoint a type must contain time data, a value,
 /// and a parameter (which is used for applying feedback).
-/// *Associated Types
-/// - TimeType: the type which represents the time of the data point.
 /// This should be trivially copyable (usually a scalar).
```
Modularius, line 6: This should be moved to after line 8

Modularius (changes requested):
In the main Cargo.toml, line 51 should be updated to
`taos = { version = "0.10.27", default-features = false, features = ["ws"] }`
as `default_features` is deprecated.

trace-to-events/src/pulse_detection/datatype/tracepoint.rs
DanNixon: Please use "Add a suggestion" it is not too clear where the comment should go...

DanNixon: Also, what does "trivially copyable" actually mean? Do you mean "implements `Copy`"?

Modularius: It means it should be cheap to copy. `Copy` is one of the required traits of `TimeType`.

Modularius (approved):

PR #309, opened 2025-02-03
Add discarded flag to digitiser aggregator spans

Modularius:
## Summary of changes

Minor code change to allow OpenTelemetry to record when a digitiser message is discarded by the aggregator for being irreparably out of chronological order.

General code check.
Has been tested with simulated and HiFi data.

### DIFF
--- a/digitiser-aggregator/src/frame/cache.rs
+++ b/digitiser-aggregator/src/frame/cache.rs
@@ -139,20 +139,20 @@ mod test {
         assert!(cache.poll().is_none());
 
         assert_eq!(cache.get_num_partial_frames(), 0);
-        let _ = cache.push(0, &frame_1, EventData::dummy_data(0, 5, &[0, 1, 2]));
+        assert!(cache.push(0, &frame_1, EventData::dummy_data(0, 5, &[0, 1, 2])).is_ok());
         assert_eq!(cache.get_num_partial_frames(), 1);
 
         assert!(cache.poll().is_none());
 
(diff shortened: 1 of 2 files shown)


digitiser-aggregator/src/frame/cache.rs
```diff
@@ -33,11 +33,11 @@ where
         digitiser_id: DigitizerId,
         metadata: &FrameMetadata,
         data: D,
-    ) {
+    ) -> bool {
```
DanNixon, line 36: To signify success or failure with no additional information prefer using `Result<(), ()>`.

DanNixon (changes requested):

DanNixon (approved):

digitiser-aggregator/src/frame/cache.rs
```diff
@@ -137,20 +139,20 @@ mod test {
         assert!(cache.poll().is_none());
 
         assert_eq!(cache.get_num_partial_frames(), 0);
-        cache.push(0, &frame_1, EventData::dummy_data(0, 5, &[0, 1, 2]));
+        let _ = cache.push(0, &frame_1, EventData::dummy_data(0, 5, &[0, 1, 2]));
```
DanNixon, line 142: Actually, is there a reason not to test the value returned here? Even if implicitly covered elsewhere.

DanNixon (changes requested):

digitiser-aggregator/src/frame/cache.rs
Modularius: That makes sense, I've added another test to test the fail case.

DanNixon (approved):

PR #316, opened 2025-02-05
Update some crates and reduce enabled features

DanNixon:
## Summary of changes

Updates some crates to their next major version.
Reduces enabled features where possible (should help dependency creep and compile times).

- code review
- test with simulated pipeline


Modularius (approved):
Tested with simulated data

PR #318, opened 2025-02-05
Set open telemetry levels via an environment variable

Modularius:
## Summary of changes

Changes the way we set OpenTelemetry levels and tidies up some spans.

- Removed `otel_level` cli option from all components.
- Modified OpenTelemetry filter layer to use environment variable "OTEL_LOG" to determine which level to filter crates/modules on.
- Removed references to the `otel` target and `target = "otel"` syntax in spans/events (This accounts for most files changed). This target is not really useful, though OpenTelemetry-only targets may be introduced in the future.
- In `trace-to-events`, functions processing individual channels moved to their own file `channels.rs`. This allows the observability level for these to be set via including `trace_to_events::channels=????` in `OTEL_LOG`. At present use `info` to include channels, and `warn/error/off` to exclude.
- Level of `spanned_root_as` functions changed to `trace` from `info`.
- Added `digitiser_id` tag to `Digitiser Event List` span in digitiser aggregator.

- General code review
- Has been tested on simulated data

### DIFF
--- a/common/src/tracer/otel_tracer.rs
+++ b/common/src/tracer/otel_tracer.rs
@@ -56,12 +56,12 @@ where
 
         let filter = match EnvFilter::builder()
             .with_default_directive(LevelFilter::OFF.into())
-            .with_env_var("OTEL_LOG")
+            .with_env_var("OTEL_LEVEL")
             .from_env()
         {
             Ok(filter) => filter,
             Err(e) => {
-                warn!("Invalid directive(s) in OTEL_LOG: {e}");
+                warn!("Invalid directive(s) in OTEL_LEVEL: {e}");
                 EnvFilter::default()
             }
         };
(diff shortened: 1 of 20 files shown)


common/src/tracer/otel_tracer.rs
```diff
@@ -63,11 +54,17 @@ where
...
             .install_batch(opentelemetry_sdk::runtime::Tokio)?;
 
-        let filter = filter::Targets::new()
-            .with_default(LevelFilter::OFF)
-            .with_target(module_name, options.level_filter)
-            .with_target("otel", options.level_filter);
-
+        let filter = match EnvFilter::builder()
+            .with_default_directive(LevelFilter::OFF.into())
+            .with_env_var("OTEL_LOG")
```
DanNixon, line 59: Is `OTEL_LOG` a standard name? If not, could `OTEL_LEVEL` or `OTEL_FILTER` be used instead as they are more descriptive.

DanNixon (changes requested):

common/src/tracer/otel_tracer.rs
Modularius: The naming mirrors the standard tracing `RUST_LOG` env variable. `OTEL_LEVEL` would be more descriptive however.

DanNixon (approved):

Issue #321, opened 2025-02-26
nexus-writer: finer grained separation of files in local directory


DanNixon:
## Problem this feature will solve

The nexus-writer should be able to handle the case where files cannot be copied to the archive directory.
Currently, it cannot cope with the situation where a copy fails.

## Description of solution

- Have two directories in the local directory: `working` and `completed`
- `working` should only ever contain one file: the one currently being written
- Any files in `working` on startup are assumed to be incomplete after a crash and are handled as they currently are
- When a file is completed, it is first moved from `working` to `completed`, after this it can be attempted to be copied to the archive
- Only after a successful copy should the file in `completed` be removed
- nexus-writer should try on startup and periodically to move any files still in `completed` to the archive, the interval should be configurable

PR #327, opened 2025-03-14
Fine grained separation of files in nexus writer

Modularius:
## Summary of changes

- Renames CLI options `file_name` and `archive_name` to `local_path` and `archive_path` respectively.
- Run files are stored in ~`local_path/temp`~ `local_path/` whilst being built up, and moved to `local_path/completed/` when finished. Failure of this file move results in component termination, and logging.
- Introduces CLI option `archive_flush_interval_sec` which tunes how often nexus-writer moves all files in `local_path/completed` to `archive_path`. `archive_flush_interval_sec` defaults to 60 (seconds) if not specified. Failure of these operations **does not** result in component termination, but warnings are emitted.
- Removed archive code from the `run.rs` and `engine.rs` to `flush_to_archive.rs` file.
- Code related to creating directory structure moved from `engine.rs` to `main.rs`
- Moved `NexusConfiguration` to `run_parameters.rs`
- The local path string is moved into `NexusSettings`, and functions which required `Some(local_path)` now requires `Some(nexus_settings)`. The use of `Option` allows unit testing.

Closes https://github.com/STFC-ICD-Research-and-Design/supermusr-data-pipeline/issues/321.

General code review.
Has been tested with simulated data.

### Breaking Changes
When running this 
- The CLI option `file_name` should be changed to `local_path`.
- The CLI option `archive_name` should be changed to `archive_path`.

### Non-Breaking Changes
- If desired, CLI option `archive_flush_interval_sec` should be set.

### DIFF
--- a/nexus-writer/src/main.rs
+++ b/nexus-writer/src/main.rs
@@ -205,7 +205,7 @@ async fn main() -> anyhow::Result<()> {
         metrics::Unit::Count,
         "Number of failures encountered"
     );
-    
+
     let run_ttl =
         Duration::try_milliseconds(args.cache_run_ttl_ms).expect("Conversion is possible");
 
--- a/nexus-writer/src/main.rs
+++ b/nexus-writer/src/main.rs
@@ -1,15 +1,17 @@
 mod nexus;
+mod flush_to_archive;
 
 use chrono::Duration;
 use clap::Parser;
+use flush_to_archive::create_archive_flush_task;
 use metrics::counter;
 use metrics_exporter_prometheus::PrometheusBuilder;
-use nexus::{NexusConfiguration, NexusEngine, NexusSettings};
+use nexus::{NexusConfiguration, NexusEngine, NexusSettings, NexusWriterResult};
 use rdkafka::{
     consumer::{CommitMode, Consumer},
     message::{BorrowedMessage, Message},
 };
-use std::{net::SocketAddr, path::PathBuf};
+use std::{fs::create_dir_all, net::SocketAddr, path::PathBuf};
 use supermusr_common::{
     init_tracer,
     metrics::{
@@ -163,6 +165,17 @@ async fn main() -> anyhow::Result<()> {
         args.archive_path.as_deref(),
     );
 
+    let mut cache_poll_interval =
+        tokio::time::interval(time::Duration::from_millis(args.cache_poll_interval_ms));
+
+    let archive_flush_task = create_archive_flush_task(&nexus_settings, args.archive_flush_interval_sec)?;
+
+    create_dir_all(nexus_settings.get_local_temp_path())?;
+    create_dir_all(nexus_settings.get_local_completed_path())?;
+    if let Some(archive_path) = nexus_settings.get_archive_path() {
+        create_dir_all(archive_path)?;
+    }
+
     let nexus_configuration = NexusConfiguration::new(args.configuration_options);
 
     let mut nexus_engine = NexusEngine::new(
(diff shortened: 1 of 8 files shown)


nexus-writer/src/main.rs
```diff
@@ -153,23 +159,30 @@ async fn main() -> anyhow::Result<()> {
...
-        nexus_settings,
-        nexus_configuration,
-    );
+    let mut nexus_engine = NexusEngine::new(Some(nexus_settings), nexus_configuration);
     nexus_engine.resume_partial_runs()?;
 
-    let mut nexus_write_interval =
-        tokio::time::interval(time::Duration::from_millis(args.cache_poll_interval_ms));
-
     // Install exporter and register metrics
```
DanNixon, line 186:
```suggestion

    // Install exporter and register metrics
```

nexus-writer/src/main.rs
```diff
@@ -192,7 +205,6 @@ async fn main() -> anyhow::Result<()> {
         metrics::Unit::Count,
         "Number of failures encountered"
     );
-
     let run_ttl =
```
DanNixon, line 208:
```suggestion

    let run_ttl =
```

DanNixon: Please avoid removing whitespace like this, it makes things that are mostly unrelated seem more related than they are.

nexus-writer/src/nexus/run_parameters.rs
```diff
@@ -8,6 +8,19 @@ use supermusr_streaming_types::{
     ecs_6s4t_run_stop_generated::RunStop, ecs_pl72_run_start_generated::RunStart,
 };
 
+#[derive(Clone, Default, Debug)]
+pub(crate) struct NexusConfiguration {
```
DanNixon, line 12: What is this used for?

DanNixon (changes requested):
On the last point in the PR description, I am pretty against having functionality in production code solely for the purpose of writing unit tests.

Could the requirement to use `Option<>` to facilitate testing be refactored out.

nexus-writer/src/nexus/run_parameters.rs
Modularius: Configuration should contain a string which is to be written to the `program_name/configuration` attribute in the Nexus File. It should describe any settings relevant to the generation of the file, e.g. it could contain the Kubernetes yaml used to create the pipeline.

It's a struct rather than a string as I wasn't sure if this data should be stored in a more structured way in the nexus file.

Modularius:
There is no `hdf5::File::default()` method, so unit tests would require creating a temp hdf5 file if we didn't include an `Option` at some point. Does it make sense to use a temp hdf5 file for every unit test?

nexus-writer/src/nexus/run_parameters.rs
DanNixon: Could this be appropriately documented in the code.

DanNixon:
> There is no `hdf5::File::default()` method, so unit tests would require creating a temp hdf5 file if we didn't include an `Option` at some point. Does it make sense to use a temp hdf5 file for every unit test?

Either:

- The functionality is not implemented in a modular enough way (if there genuinely is something that can be tested in the absence of files), or
- The tests are not appropriate as they are not using files.

What are the impacted tests?

Modularius:
It's the tests in `engine.rs`.

They test `NexusEngine`'s functionality (e.g. processing runstart/runstop, timestamps) without the overhead and potential errors that come from disk I/O.

This way of doing things does feel a bit hacky, I can refactor it.

DanNixon:
@Modularius there are a couple of conflicts to resolve.

DanNixon (approved):

Issue #337, opened 2025-04-23
Add docstrings to Nexus Writer Component


Modularius:
## Description of task

Add docstrings to document all parts of the nexus-writer component. All functions/methods/structs/enums/etc. should be documented to explain what they do and what failure modes are possible.

PR #347, opened 2025-05-15
Add docstrings to the nexus writer

Modularius:
## Summary of changes

Adds docstrings to every module of the nexus writer (plus a few changes to other components before it was decided to do them separately).

No changes to the code have been made.

## Still to do
`crate::hdf5_handlers` is incomplete as some refactoring should be done to simplify.

- Check no actual code has been modified.
- Check the descriptions make sense, are instructive, and correct.
- Check spelling and grammar (everything should end in a full stop, except for some lists).
- Check `cargo doc --no-deps` produces expected documentation.

Refs [Add docstrings to Nexus Writer Component](https://github.com/STFC-ICD-Research-and-Design/supermusr-data-pipeline/issues/337).


nexus-writer/src/hdf5_handlers/dataset.rs
```diff
@@ -13,10 +15,12 @@ impl HasAttributesExt for Dataset {
         Ok(attr)
     }
 
+    /// This should be a provided method.
```
DanNixon, line 18: Functions in trait `impl`s should not be documented (they inherit the documentation from the trait).

nexus-writer/src/hdf5_handlers/dataset.rs
```diff
@@ -1,3 +1,5 @@
+//! This module implements the [DatasetExt] and [HasAttributesExt] traits for
```
DanNixon, line 1: A literal list of what is in the module is not overly helpful, if there is not a higher level description that is relevant then it is preferable to omit this.

nexus-writer/src/hdf5_handlers/error.rs
```diff
@@ -277,14 +347,18 @@ impl From<TryFromIntError> for NexusHDF5Error {
     }
 }
 
-/// Used to allow errors which can be convertex to NexusHDF5Errors to be
-/// appended with hdf5 paths
+/// Used to allow errors which can be convertex to [NexusHDF5Error] to be
+/// appended with hdf5 paths. This is intended to be implemented by [Result<_,E>]
+/// where `E: Error + Into<NexusHDF5Error>`.
```
DanNixon, line 352:
```suggestion
```

This is already said in the definition.

nexus-writer/src/hdf5_handlers/error.rs
```diff
@@ -277,14 +347,18 @@ impl From<TryFromIntError> for NexusHDF5Error {
...
 pub(crate) trait ConvertResult<T, E>
 where
     E: Error + Into<NexusHDF5Error>,
 {
+    /// Implementation should convert the error to [NexusHDF5Error] type and add the path of `group` to it.
```
DanNixon, line 357:
```suggestion
    /// Converts the error to [NexusHDF5Error] type and add the path of `group` to it.
```

Document trait functions in the context of being read by the user, not the trait implementer. Otherwise you end up with two similar but different context in the way docs are written.

nexus-writer/src/hdf5_handlers/group.rs
```diff
@@ -8,27 +10,56 @@ use hdf5::{
 };
 
 impl HasAttributesExt for Group {
+    /// Creates a new attribute, with name as specified.
+    /// # Parameters
+    ///  - attr: name of the attribute to add.
+    /// # Error Modes
```
DanNixon, line 16: Header with no content.

nexus-writer/src/nexus_structure/entry/mod.rs
```diff
@@ -52,37 +53,65 @@ mod labels {
 }
 
 // Values of Nexus Constant
+/// The instrument definition file version, currently "2".
 const IDF_VERSION: u32 = 2;
+/// Muon Timeseries Database.
 const DEFINITION: &str = "muonTD";
+/// This may be changed to something more widely agreed upon. Possibly statically scraped from elsewhere.
 const PROGRAM_NAME: &str = "SuperMuSR Data Pipeline Nexus Writer";
+/// This should probably be taken statically from "cargo.toml" or something.
```
DanNixon, line 62: May as well just do that rather than documenting it. See https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates

nexus-writer/src/nexus_structure/entry/mod.rs
```diff
@@ -52,37 +53,65 @@ mod labels {
 }
 
 // Values of Nexus Constant
+/// The instrument definition file version, currently "2".
 const IDF_VERSION: u32 = 2;
+/// Muon Timeseries Database.
```
DanNixon, line 58: This does not shed any light on what this const is for.

DanNixon (changes requested):
I am only a few files in but I have noticed a few issues of the same type happening quite often:

- If the comment says nothing more than the code does, then it is just noise and should not be added
- I don't immediately see what value a lot of the "Error Modes" sections bring, that should be sufficiently obvious from the return type unless the function is expected to panic

README.md
```diff
@@ -3,3 +3,143 @@
 [![CI](https://github.com/DanNixon/supermusr-data-pipeline/actions/workflows/ci.yml/badge.svg)](https://github.com/DanNixon/supermusr-data-pipeline/actions/workflows/ci.yml)
 
 See [docs](./docs).
+
+
+## Nexus Writer
```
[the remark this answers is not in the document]
Modularius: No, this was meant to be removed before pushing.

nexus-writer/src/nexus_structure/entry/mod.rs
Modularius: I wanted to keep this PR from editing any of the code. I'll change this in another PR.

Modularius:
I've made the changes requested in the Nexus Writer component, but I've left the others as I'll work on them in a separate PR. I'm going to alter the `hdf5_handler` module, so some docstrings in the `impl` blocks have been left to be removed later.

DanNixon:
In the first pass I somewhat missed that there is documentation included that is not for `nexus-writer`, why is that the case?

Modularius:
The PR was originally going to be for all components. I thought I may as well leave the docs that I started in the other components and finish them in another PR

PR #349, opened 2025-05-20
Enable zstd compression

DanNixon:
## Summary of changes

Provide the zstd library and enable zstd compression by default in all components that interact with Kafka.

This is partly experimental to see what the effects are on compute requirements and network throughput are for data "inside" the pipeline.

- Run a demo pipeline, using the container images, with and without these changes and see that the same output is still obtained.
- Verify that the data produced is actually compressed (can be done via Redpanda Console).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -188,7 +188,7 @@ version = "0.29.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "61b1d86e7705efe1be1b569bab41d4fa1e14e220b60a160f78de2db687add079"
 dependencies = [
- "bindgen",
+ "bindgen 0.69.5",
  "cc",
  "cmake",
  "dunce",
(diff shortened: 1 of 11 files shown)


DanNixon:
> Tested the pipeline with compression enabled, but it ran too slowly to test its correctness. The lag for the trace-to-events consumer group increased to the point it couldn't process the messages in time for the runs.

I assume this was testing with actual data?
If so, that should be avoided here as 1) there is insufficient network bandwith between R55 and R89 to run the pipeline correctly (we almost never want to be streaming trace data out of the instrument cabins for any reason), 2) this would not be an accurate comparison as you will not be decompressing anything in trace-to-events, 3) testing at the full pipeline bandwidth is not actually that important, only relative speeds are really useful, 4) the simulator should also be tested (if other things work, it also should, but it can't hurt to be included).

> If the slowdown is an issue with the compression, I would imagine it occurs because the trace messages are stored uncompressed on the broker, so need to be compressed on the broker when requested.

This is not how compression works in Kafka.
In the simple producer-topic-fanout case, it is the producer that handles the compression, the broker just stores what it receives and the clients must handle any compression in what they pull from the broker.
Clients do not request data to be compressed with a specific algorithm.

DanNixon:
Where are you getting the message sizes from? Are you certain they are not showing the uncompressed payload size?

DanNixon:
IIRC, that will be showing the uncompressed size.

Right now, it is not too critical to measure the compression ratio here as the biggest saving if any will be in the trace data, which is hopefully going to be benchmarked this week.

Even uncompressed, the "in-pipeline" data volume is very small in comparison.

DanNixon:
Main question is: do the container images work?

PR #387, opened 2025-08-19
Prefix metric names with `muon_data_pipeline_`

DanNixon:
## Summary of changes

Adds "muon_data_pipeline_" prefix to all common metric names.
Aids in finding metrics in the collection of other things collected by Prometheus.

- Run any component that exposes metrics
- `curl` the metrics endpoint
- See that metrics names have the expected prefix

(or code review is realistically enough)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4649,6 +4649,7 @@ name = "supermusr-common"
 version = "0.1.0"
 dependencies = [
  "clap",
+ "const_format",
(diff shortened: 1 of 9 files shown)


DanNixon:
> Trace to Events has an additional metric, whose name is defined on line 46 of main.rs. You could either edit this in-place, or move it to the common crate.

Good spot.

Issue #403, opened 2025-09-11
Trace viewer cannot be served from a path


DanNixon:
## Description of the problem

Try to deploy the trace viewer at a path under a domain, [it does not work](https://ndw2982.isis.cclrc.ac.uk/trace-viewer).

It should either use a relative path when making API calls or allow the expected URL it is reached via to be set on the command line.

PR #405, opened 2025-09-13
Server path

Modularius:
## Summary of changes

- Adds optional `server-path` parameter for the command line.
- Modifies links and path to web resources.

Test on MuSR with `--server-path trace-viewer` added to command line.

Closes [Trace viewer cannot be served from a path](https://github.com/STFC-ICD-Research-and-Design/supermusr-data-pipeline/issues/403).

### DIFF
--- a/trace-viewer/src/app/mod.rs
+++ b/trace-viewer/src/app/mod.rs
@@ -29,7 +29,16 @@ pub(crate) struct TopLevelContext {
     client_side_data: ClientSideData,
 }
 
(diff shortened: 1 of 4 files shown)


DanNixon (changes requested):
Suggestions avoid having to handle `Option` where not strictly required to do so.

DanNixon:
@Modularius Is it intentional that the default with your changes is "" rather than "/"?

DanNixon (approved):

PR #410, opened 2025-09-30
Add extra search option in trace viewer

Modularius:
## Summary of changes

Adds a new search option (dragnet searching, name is a placeholder), which uses binary search to locate a given timestamp, jumps back the specified number of messages then searches forwards for the specified number of messages.

This is intended as a blunt instrument mode, which does not rely on messages being in chronological order. This allows the tool to be used to help diagnose issues with messages appearing out of order in the broker.

The layout on the webpage is confusing, but this will be updated in a future PR.

General code review.
Please review after https://github.com/STFC-ICD-Research-and-Design/supermusr-data-pipeline/pull/409 is merged.


DanNixon:
I did actually mean rebase here (not merge, running `git merge` is almost never what you want to do). You need to remove the commits that were part of the other PR.
(technically Git would resolve this at merge time, but it makes reviewing more difficult as it shows changes from both PRs)

PR #433, opened 2025-11-04
write event data to `detector_1_events` instead of `detector_1`

rerpha:
## Summary of changes

I think in a traditional nexus file events are put in the `detector_1_events` group - this changes the nexus-writer to do the same. Also removes a stray string in favour of a constant. event mode files normally have a `detector_1` group but for retrospectively histogrammed data. 

No testing needed really - the CI should be enough.


DanNixon:
@Modularius Was there a reason for the original name that I might be missing? I'd agree `detector_1_events` would be more inline with what I'd expect.

Modularius:
I just took the name from existing NeXus files (these mostly were *not* event files), I'm not aware of any reason why it has to be that name.

PR #485, opened 2026-04-22
Update rustls-webpki

Modularius:
## Summary of changes

Add new advisory to ignore list of `deny.toml`

General review. No code changes.


DanNixon:
There is a [patched version that is not affected](https://rustsec.org/advisories/RUSTSEC-2026-0104.html), can the version of `rustls-webpki` just be upgraded?

Modularius:
Ah yes, in this case can any advisories producing a "advisory was not encountered" warning be removed from deny.toml?

DanNixon:
> Ah yes, in this case can any advisories producing a "advisory was not encountered" warning be removed from deny.toml?

Yep.

DanNixon (approved):

PR #498, opened 2026-06-19
Aggregator: force non-repeating digitiser ids

### LINKED ISSUE #338: Digitiser Aggregator check
## Problem this feature will solve

If the user runs the `digitiser-aggregator` with a non-increasing sequence of digitiser ids or a sequence that contains repetitions, the aggregator will never recognise frames as complete. Currently no warning or error occurs if the user does this.

## Description of solution

The aggregator should panic if the user provides a non-increasing or repeating sequence of digitiser ids.

Modularius:
## Summary of changes

Adds validation step that checks the user has passed digitiser ids as a ~~strictly increasing~~ non-repeating sequence, and bails otherwise.

Forcing the user to adhere to this format rather than fixing it silently is more likely to catch user cli typos.

Check `cargo run -p digitiser-aggregator -- --broker 0 --group 1 --input-topic a --output-topic b -d0,1,2,3,4,6,8` works,
and that `cargo run -p digitiser-aggregator -- --broker 0 --group 1 --input-topic a --output-topic b -d0,1,2,3,4,6,8` works,
and `cargo run -p digitiser-aggregator -- --broker 0 --group 1 --input-topic a --output-topic b -d0,1,2,4,3,6,6` doesn't.

Closes https://github.com/ISISNeutronMuon/digital-muon-pipeline/issues/338.

### DIFF
--- a/digitiser-aggregator/src/frame/cache.rs
+++ b/digitiser-aggregator/src/frame/cache.rs
@@ -1,6 +1,9 @@
 //! Defines the cache stores frames as they are assembled from digitiser messages.
 use super::{AggregatedFrame, RejectMessageError, partial::PartialFrame};
-use crate::{data::{Accumulate, DigitiserData}, frame::FrameCacheError};
+use crate::{
+    data::{Accumulate, DigitiserData},
+    frame::FrameCacheError,
+};
 use chrono::{DateTime, Utc};
 use digital_muon_common::{
     DigitizerId, record_metadata_fields_to_span, spanned::SpannedAggregator,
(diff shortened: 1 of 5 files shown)


digitiser-aggregator/src/frame/cache.rs
```diff
@@ -33,13 +36,27 @@ where
...
-            expected_digitisers,
-            latest_timestamp_dispatched: None,
-            frames: Default::default(),
+    /// Note that the function returns an error if `expected_digitisers` has any repetitions.
+    pub(crate) fn new(
+        ttl: Duration,
+        mut expected_digitisers: Vec<DigitizerId>,
+    ) -> Result<Self, FrameCacheError> {
+        expected_digitisers.sort();
+        let validation_fn = |pair: &[DigitizerId]| {
```
DanNixon: Move this back to being `clap` validation and just use the itertools function (https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.duplicates).

DanNixon (changes requested):

digitiser-aggregator/src/frame/cache.rs
Modularius: I've used `itertools`, but clap doesn't have a native method for requiring unique values.

DanNixon (approved):

PR #499, opened 2026-06-19
Optimise trace writer

Modularius:
## Summary of changes

Changes format of the trace writer's output, whilst adding attributes to indicate this, allowing trace-reader to manage differing formats.

- Stores timestamps as ns since epoch instead of strings.
- Stores all channel trace data in a single 3D dataset extensible in the first axis only.
- Rejects any trace messages with incorrect number of channels or trace sizes. These are determined from the first trace message received.
- Adds test_assets folder with dat2 fb message and hdf5 files used in testing. The test reads the fb message, writes it to a memory only hdf5 file and compares it with the asset hdf5.

General code review.


trace-writer/src/digitiser_data.rs
```diff
@@ -0,0 +1,317 @@
...
+    sizes: &[usize; 3],
+    trace_size: usize,
+) -> bool {
+    if channels.clone().any(|c| c.voltage().is_none()) {
+        warn!("Missing channel voltages.");
+        return false;
+    }
+    if channels
+        .clone()
+        .any(|c| c.voltage().map(|v| v.len() != trace_size).unwrap_or(true))
```
DanNixon: A change in acquisition length is not something that should cause data to be discarded.

trace-writer/src/digitiser_data.rs
```diff
@@ -0,0 +1,317 @@
...
+        .new_dataset::<T>()
+        .shape(SimpleExtents::fixed(vec![size]))
+        .create(name)
+}
+
+/// Checks that the existing traces dataset has the appropriate shape for the current message.
+/// # Parameters
+/// - channels: iterator over the current message's channels.
+/// - sizes: the dimensions of the hdf5 traces dataset.
+/// - trace_size: the size of the traces of the current message.
```
DanNixon: Isn't `trace_size` derivable from `channels`?

DanNixon (changes requested):

trace-writer/src/digitiser_data.rs
Modularius: Should the writer be able to accommodate different acquisition lengths, or can we just clip longer traces? This affects how the traces can be stored internally.

Modularius: Ideally, I want to store all trace data for a digitiser in a single 3D dataset of `u16` of dims `[number of messages, number of channels, trace size]`, resizing the first dimension each time a new message comes in.
If we accommodate arbitrary trace sizes, we could either resize the 3rd dimension to the largest value (which might cause an unacceptable amount of reallocation), or store data as a 2D dataset of `VarLenArray<u16>` with `[number of messages, number of channels]`.

I can't find much in the way of advice on the best layout. I'd be in favour of allowing the user to choose how it's stored.

DanNixon (approved):

PR #505, opened 2026-07-02
Upgrade anyhow to fix cargo deny issue

Modularius:
## Summary of changes

Pins anyhow version in `cargo.toml` to fix issue with `cargo deny check`. A bug in `anyhow 1.0.100` may affect crates using it such as `leptos`.

Ensure `cargo deny check` runs.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ edition = "2024"
 [workspace.dependencies]
 actix-files = "0.6"
 actix-web = { version = "4", features = ["macros"] }
+anyhow = "1.0.103"
 assert_approx_eq = "1.1.0"
 cfg-if = "1"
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -22,6 +22,7 @@ edition = "2024"
 [workspace.dependencies]
 actix-files = "0.6"
 actix-web = { version = "4", features = ["macros"] }
+anyhow = "1.0.103"
```
DanNixon: Do we actually need this? Was `cargo update anyhow` not enough?

DanNixon (changes requested):

Cargo.toml
Modularius: You're right, I'll remove it.

DanNixon (approved):

